PR #38, opened 2016-12-07
Use nested eigen submodule

oliverlee:


oliverlee:
@mickvangelderen Want to review?

mickvangelderen:
I've read the changes but still cannot execute them.

oliverlee:
It doesn't build? You may need to clear the Eigen CMake cache variable if
you aren't using a new build directory.

PR #48, opened 2016-12-14
Fix pitch minimum value

oliverlee:


oliverlee:
This looks okay from a few minutes of testing.

PR #49, opened 2016-12-14
Plot dt in pose plots

oliverlee:
Calculate time difference between pose timestamp values and add as a
subplot to pose plots. This field has millisecond resolution.


oliverlee:
screen shot 2016-12-14 at 15 19 50

PR #52, opened 2017-01-06
Add bicycle simulation class

oliverlee:
The Bicycle class in sim namespace is a template class that simulates a
bicycle model with an observer model. This class:
- incorporates fields necessary for visualization, such as wheel angle
- provides a single interface for using different bicycle models
- allows simulation of dynamics and kinematics separately


oliverlee:
@bazilinskyy could use a review

oliverlee:
details to be worked out:
 - [ ] Estimation of handlebar feedback torque. Will use of estimate state and (possibly noisy) input vectors be sufficient?
 - [ ] What to do at low/zero forward velocity?
 - [ ] What to do when the bicycle falls over (abs(roll) > 90 deg)?
 - [ ] Quantization of bicycle forward velocity to reduce frequency of discrete-time state space calculation? Although both dynamics and kinematics are calculated in the same loop so perhaps this should postponed until multiple loops are running.

PR #56, opened 2017-01-12
Fix: Use size_t instead of uint8_t to specify buffer sizes

mickvangelderen:


oliverlee:
Thanks for your contribution.

PR #76, opened 2017-02-21
Add python record plots

oliverlee:


oliverlee:
Plots should contain state information show differences between state estimate vs recorded data (particularly steer angle). Some sample protobuf logs with and without handlebar torque enabled.
[logs.zip](https://github.com/oliverlee/phobos/files/790565/logs.zip)

oliverlee:
Plot of measured steer angle vs. Kalman filter estimated steer angle shows a large difference:
screen shot 2017-02-21 at 22 38 29

oliverlee:
This is okay for now but better plotting tools are still necessary.

PR #81, opened 2017-02-22
Tune kalman covariances

oliverlee:


oliverlee:
This seems to work for now but may need to be adjusted in the future.

PR #88, opened 2017-02-22
WIP Fix roll saturation states

oliverlee:


oliverlee:
We are removing roll saturation all together. See issue https://github.com/oliverlee/phobos/issues/103

PR #95, opened 2017-02-24
Fix steer torque calculation

oliverlee:


oliverlee:
@mickvangelderen

scripts/calculate_handlebar_inertia.py
```diff
@@ -0,0 +1,67 @@
...
+if __name__ == '__main__':
+    if len(sys.argv) < 2:
+        print('Usage: {} <log_file>\n'.format(__file__))
+        print('    <log_file>\tFile containing serialized and framed ' +
+              'protobuf messages')
+        sys.exit(1)
+
+    messages = load_messages(sys.argv[1])
+    # ignore first sample as it transmitted before the simulation loop
+    records = get_records_from_messages(messages)[1:400]
```
mickvangelderen: Maybe you should mention what kind of samples you are looking for (not close to 0) too?

projects/inc/saconfig.h
```diff
@@ -56,4 +56,7 @@ constexpr float STEER_ASSEMBLY_INERTIA_WITH_WEIGHT = 0.1942f; /* with 1 kg weigh
 constexpr float STEER_ASSEMBLY_INERTIA_WITHOUT_WEIGHT = 0.0828f; /* without weight plates */
 constexpr float STEER_ASSEMBLY_INERTIA = STEER_ASSEMBLY_INERTIA_WITH_WEIGHT; /* default configuration */
 
+/* moment of inertia of the handlebars and steering column above the torque sensor, about the steer axis */
+constexpr float HANDLEBAR_INERTIA = 0.11889455460271312f; /* kg-m^2 */
```
mickvangelderen: This definition makes the value seem more precise than it actually is.

mickvangelderen:
Some of the commits to a single file might be better off when squashed together so that the code is "complete" at every commit.

scripts/calculate_handlebar_inertia.py
oliverlee: Output value was the same for [1:400] and [1:] so I'll just change it.

mickvangelderen: What if you take [400:]?

oliverlee:
@mickvangelderen Embedded code squashed.

scripts/calculate_handlebar_inertia.py
oliverlee: The script just calculates a number and probably will be run infrequently. I think we can expect the user to change indices as necessary.

PR #98, opened 2017-02-24
Add multiplatform debug statements

mickvangelderen:


inc/debug.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#if defined(CH_CFG_TIME_QUANTUM)
+// We are compiling for ChibiOS
+#include "osal.h"
+#define debugAssert(condition, remark) osalDbgAssert(condition, remark)
+#define debugCheck(condition) osalDbgCheck(condition)
+#else
+// We are not compiling for ChibiOS
+#include <assert.h>
```
oliverlee: #include \<cassert\>
https://stackoverflow.com/questions/5197136/why-is-there-a-c-prefix-in-the-c-version-of-c-libs

inc/debug.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#if defined(CH_CFG_TIME_QUANTUM)
+// We are compiling for ChibiOS
+#include "osal.h"
+#define debugAssert(condition, remark) osalDbgAssert(condition, remark)
+#define debugCheck(condition) osalDbgCheck(condition)
```
oliverlee: I prefer lowercase_with_underscores for function names.

src/packet/frame.cc
```diff
@@ -12,12 +10,12 @@
...
-#if defined(CH_CFG_TIME_QUANTUM) // ChibiOS CFG options are always defined by CMake
-    osalDbgCheck(source != nullptr);
-    osalDbgCheck(dest != nullptr);
-    osalDbgAssert(source_byte_size <= packet::frame::COBS_MAX_SIZE_PAYLOAD,
-            "source data exceeds allowed frame algorithm size");
-#endif // defined(CH_CFG_TIME_QUANTUM) // ChibiOS CFG options are always defined by CMake
+    debugCheck(source != nullptr);
+    debugCheck(dest != nullptr);
+    debugAssert(
+      source_byte_size <= packet::frame::COBS_MAX_SIZE_PAYLOAD,
```
oliverlee: tabs/indents are always 4 spaces _unless_ aligning with parenthesis

inc/debug.h
```diff
@@ -0,0 +1,13 @@
...
+    // We are compiling for ChibiOS
+    #include "osal.h"
+    #define debug_assert(condition, remark) osalDbgAssert(condition, remark)
+    #define debug_check(condition) osalDbgCheck(condition)
+#else
+    // We are not compiling for ChibiOS
+    #include <cassert>
+    #define debug_assert(condition, remark) assert(condition)
+    #define debug_check(condition) assert(condition)
+#endif
```
oliverlee: i generally like to do:
```cpp
#endif // defined(CH_CFG_TIME_QUANTUM) 
```
as it makes it easier to track if multiple compiler directives are used

PR #99, opened 2017-02-24
Add bicycle kinematic model

oliverlee:
- [x] test with hardware

### DIFF
--- a/projects/inc/messageutil.h
+++ b/projects/inc/messageutil.h
@@ -18,18 +18,38 @@ namespace message {
     void set_bicycle_canonical(BicycleModelMessage* pb, const bicycle_t& b);
     void set_bicycle_discrete_time_state_space(BicycleModelMessage* pb, const bicycle_t& b);
 
-    template <typename kalman_t>
-    void set_symmetric_output_matrix(SymmetricOutputMatrixMessage* pb,
-            const typename kalman_t::measurement_noise_covariance_t& m);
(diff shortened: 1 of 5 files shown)


projects/inc/messageutil.h
```diff
@@ -18,18 +18,38 @@ namespace message {
...
-    void set_kalman_gain(BicycleKalmanMessage* pb,
-            const kalman_t& k);
+    /* functions for different observer variants */
+    template <typename observer_t>
+    typename std::enable_if<std::is_same<observer_t, typename observer::Kalman<typename observer_t::model_t>>::value, void>::type
+    set_symmetric_output_matrix(SymmetricOutputMatrixMessage* pb, const typename observer_t::measurement_noise_covariance_t& m);
+
+    template <typename observer_t>
+    typename std::enable_if<!std::is_same<observer_t, typename observer::Kalman<typename observer_t::model_t>>::value, void>::type
+    set_symmetric_output_matrix(SymmetricOutputMatrixMessage* pb, const typename observer_t::measurement_noise_covariance_t& m);
```
mickvangelderen: There seems to be no difference between this line and line 24. Is that correct?

oliverlee:
Hardware tests show the Kalman message entries only appear flimnap_whipple and not in flimnap_kinematic. Quick tests also indicate (subjectively) that feedback torque is correct (using HandlebarStatic).

PR #102, opened 2017-02-26
Add Whipple model simulation script

oliverlee:


oliverlee:
Probably not important since these are just used for plotting the data.

PR #106, opened 2017-02-28
:sparkles: Add tested COBS implementation

mickvangelderen:
code is tested, but the testing framework is.. not there

:hammer: Decide if the shortcut approach is a good idea.

:mag: Decide if we want to append a 0 to the output. I think splitting the input stream into decodable chunks (packets) is not the job of cobs::encode/cobs::decode.

:mag: Look into a streaming implementation.


mickvangelderen:
@oliverlee I do not intend to merge this, I just want your feedback on how to go from here.

oliverlee:
Ideally the test binary should be built with CMake. For now, we can add a post-build statement to Travis so that the test is run, but the return value needs to be non-zero and ideally the test error would be printed so we don't need to re-run the test locally but simply look at the Travis log.

To address the first post:

> 🔨 Decide if the shortcut approach is a good idea.
I'll have to look at this later.

> 🔍 Decide if we want to append a 0 to the output. I think splitting the input stream into decodable chunks (packets) is not the job of cobs::encode/cobs::decode.
I think it should. What's the use case for using cobs encode without appending a 0 to the output? A separate packet framing function would need to know that the data is encoded with cobs and then append a 0.

> 🔍 Look into a streaming implementation.
This algorithm is fairly inexpensive computationally. Would streaming be for extremely large data objects? The largest object we encode (currently) has a maximum size of 700 bytes or so.
It seems this would require most knowledge of how the data is to be streamed and would be tied to an OS for an efficient implementation. I don't see the benefit over copying bytes.

tests/cobs_test.cc
```diff
@@ -0,0 +1,145 @@
...
+Throw exceptions so we can catch them with gdb (catch catch).
+*/
+void throw_assert(bool condition, const char * const remark) {
+  if (condition == false) throw remark;
+}
+
+/*
+Fill a byte buffer with a given value.
+*/
+void fill(uint8_t * const b, size_t l, uint8_t v) {
```
mickvangelderen: Fair, using memset now. Kept the argument order of fill because that is consistent with passing the length of a buffer immediately after the buffer.

tests/cobs_test.cc
```diff
@@ -0,0 +1,145 @@
...
+  for (size_t i = 0; i < l; i++) {
+    b[i] = v;
+  }
+}
+
+/*
+Check that a byte buffer contains only a given value.
+*/
+void cmpb(uint8_t * const b, size_t l, uint8_t v) {
```
mickvangelderen: compare buffer? renamed to mem_is.

tests/cobs_test.cc
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+int main() {
+  const uint8_t X = 0xdd; // non-zero payload
+  const uint8_t F = 0xcc; // fill
+
+  // Encode the empty packet.
+  {
+    uint8_t b1[1000]; fill(b1, sizeof(b1), X);
```
mickvangelderen: This is a by product of not being able to declare an array filled with a given value. 

If you would accept

```cpp
list<uint8_t> b = new list<uint8_t>(1000).fill(X);
```

Then the current way of writing it should be fine too.

tests/cobs_test.cc
```diff
@@ -0,0 +1,145 @@
+#include "../inc/cobs.h"
+
+/*
+Throw exceptions so we can catch them with gdb (catch catch).
+*/
+void throw_assert(bool condition, const char * const remark) {
+  if (condition == false) throw remark;
```
mickvangelderen: I've found it much more powerful and less tedious to simply debug your tests than to try and give all information you could possibly need to debug the problem in the form of an error message + stack trace. 

This is a simple solution to simple problem. I can spend time on adopting a testing framework but that is not a priority right?

tests/cobs_test.cc
oliverlee: Or std::fill if you aren't working with an array of bytes.

tests/cobs_test.cc
oliverlee: In general, I place each statement on a separate line so that is easier to diff (sometimes) but mostly for stepping through a debugger. Note that these are two separate instructions: an initialization followed by a set.

I keep with that guideline so what I would do is:
```cpp
uint8_t b1[1000];
fill(b1, sizeof(b1), X);
```
even though this is a simple case in order to be consistent.
I'd also use std::array (or std::vector) instead of std::list to allow fast random access.

tests/cobs_test.cc
oliverlee: Sure using a debugger with tests is easier and more powerful. The problem is when a change is implemented in the future that causes the test to fail. In that case, you want to be able to look at the CI log and quickly see what is failing and why. It doesn't have to be an extensive.

The problem with throwing exceptions is that it usually halts your test framework (dunno maybe this has changed) so you can't get a report of _all_ the tests.

mickvangelderen: Thank you for the clarification but you have not told me whether or not doing this is a priority.

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+/*
+Consistent Overhead Byte Stuffing (COBS) removes a specific value from a
+list of values. The removed value can then, for example, be used as a
+frame marker.
+
+Refer to: https://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing
+
+COBS considerations:
+
+1. Configurable marker byte value. The current implementation assumes
```
oliverlee: How exactly does this work if the marker byte is not 0? The 0 implementation is easy to understand because you simply use overhead and altered bytes that point to the next zero location. What if the marker byte is 0x03 and you need to point to a location 3 bytes away?

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+
+4. Append the marker byte to the encoded result. The current
+implementation does append the marker byte.
+
+Here are some interesting cases. x is a non-zero byte. 0 is
+a zero byte. 123x denotes 123 consecutive x's.
+
+decoded, length -> encoded, length, note
+       , 0      -> 0      , 1     , short encoding
+       , 0      -> 1|0    , 2     , standard encoding
```
oliverlee: There is no data to encode. Why does this return anything?

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+x      , 1      -> 2|x|0  , 3     ,
```
oliverlee: This and the line below should be swapped, with a note added that x != 0

oliverlee: whoops missed the note on line 30

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+The current implementation applies this reduced encoding.
+
+3. If the encoding results in { n, ..., x, 0 } where n is an offset and
+x > n, encode it as { x, ..., 0 }. The standard implementation would run
+into an decoding error because x > n is bigger than the remaining number
+of bytes. The current implementation does not apply this reduced
+encoding.
+
+4. Append the marker byte to the encoded result. The current
+implementation does append the marker byte.
```
oliverlee: I don't see a reason not to append the marker byte. It will also make the examples on lines 34-58 confusing because you append the marker byte there.

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+decoded, length -> encoded, length, note
+       , 0      -> 0      , 1     , short encoding
+       , 0      -> 1|0    , 2     , standard encoding
+x      , 1      -> 2|x|0  , 3     ,
+0      , 1      -> 1|1|0  , 3     ,
+x|x    , 2      -> 3|x|x|0, 4     ,
+x|0    , 2      -> 2|x|1|0, 4     ,
+0|x    , 2      -> 1|2|x|0, 4     ,
+0|0    , 2      -> 1|1|1|0, 4     ,
+ encoded       , length, note
```
oliverlee: are you missing a column descriptor here?

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+frame marker.
+
+Refer to: https://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing
+
+COBS considerations:
+
+1. Configurable marker byte value. The current implementation assumes
+0x00.
+
+2. If the encoding results in { ..., 0, 1 }, encode it as { ..., 0 }.
```
oliverlee: Is this supposed to be reversed? 
```
{..., 1, 0} and not {..., 0, 1}
```

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+0      , 1      -> 1|1|0  , 3     ,
+x|x    , 2      -> 3|x|x|0, 4     ,
+x|0    , 2      -> 2|x|1|0, 4     ,
+0|x    , 2      -> 1|2|x|0, 4     ,
+0|0    , 2      -> 1|1|1|0, 4     ,
+ encoded       , length, note
+252x|0 , 253    -> 253|252x|1|0  , 255   ,
+253x   , 253    -> 254|253x|0    , 255   ,
+253x|0 , 254    -> 254|253x|1|0  , 256   ,
+254x   , 254    -> 255|254x|0    , 256   , short encoding
```
oliverlee: Is short encoding the same as what you refer to as reduced encoding on line 18? If so, use the same name/words.

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+
+Refer to: https://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing
+
+COBS considerations:
+
+1. Configurable marker byte value. The current implementation assumes
+0x00.
+
+2. If the encoding results in { ..., 0, 1 }, encode it as { ..., 0 }.
+The current implementation applies this reduced encoding.
```
oliverlee: I don't think this is a good idea. While we save one byte (possibly), it makes the calculation of encoded/decoded size more complicated.

src/cobs.cc
```diff
@@ -0,0 +1,124 @@
...
+    It could be implemented as:
+
+    max_enc_len(len) {
+        return len + ceil(len/254) + 1
+    }
+
+    To get rid of the ceil we can abuse integer division:
+
+    max_enc_len(len) {
+        return len + (len + 254)/254 + 1
```
oliverlee: this should be:
```
len + (len + 254 - 1)/254 + 1
```

src/cobs.cc
```diff
@@ -0,0 +1,124 @@
...
+    }
+    */
+    size_t max_encoded_length(size_t decoded_length) {
+        return decoded_length + (decoded_length + 0xfe)/0xfe + 1;
```
oliverlee: return decoded_length + (decoded_length + 0xfe - 1)/0xfe + 1;

src/cobs.cc
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /*
+    The smallest encoded size is obtained when the input contains no
+    zeros more than 254 bytes apart. In this case the overhead is only 2
+    bytes: the first zero offset and the frame marker. Thus the maximum
+    decoded length is the length minus 2. The empty packet encodes as a
+    single 0.
```
oliverlee: Again, what's the point in encoding an empty packet?

tests/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+project(phobos_tests)
+
+add_executable(test_cobs
+  test_cobs.cc
+  ../src/cobs.cc
+)
+
+target_include_directories(test_cobs PRIVATE ../inc)
+
+install(TARGETS test_cobs RUNTIME DESTINATION bin)
```
oliverlee: Do really want to install this?

oliverlee:
Haven't finished reviewing but added some comments.

inc/cobs.h
mickvangelderen: I think

> It just works (tm)

`[0, 1, 2, 3, 4, 5, 3, 6, 7, 8]` would be encoded as `[(4), 0, 1, 2, {3}, 4, 5, {4}, 6, 7, 8, 3]`. (n) means an offset of n where a 3 *is not* inserted. {n} means an offset of n where a 3 *is* inserted.

700 non 3 bytes will be encoded as `[(255), 254 non 3, (255), 254 non 3, (193), 192 non 3, 3]`. 

The alternative is to simply XOR the input with 0x03 before encoding.

inc/cobs.h
mickvangelderen: Because the presence of a packet is also information, even if the contents are empty. If you don' t want to send empty packets then don' t call cobs::encode.

inc/cobs.h
mickvangelderen: The reason I put it into the considerations is that I don' t see any downsides to not appending the marker. It is a decision that has to be made and people looking for a cobs implementation will want to know that I at least thought about it.

inc/cobs.h
mickvangelderen: good point

inc/cobs.h
mickvangelderen: yeah this is not right.

inc/cobs.h
mickvangelderen: I am retarded

mickvangelderen: Of course you'd have to define the meaning of the offsets slightly differently:

offset `n` | followed by
---|---
0 - 2 | `n` data bytes and a 3
4 - 254 | `n - 1` data bytes and a 3
255 | `n - 1` data bytes

oliverlee: What does it mean to have an offset of 0 in this case?

inc/cobs.h
oliverlee: Fair enough. However, this encoding scheme is primarily being used for unambiguous packet framing for serial data. This means the use case will practically always require appending the marker byte. Again, what is the use case for doing a COBS encode without appending the marker byte? There are many other encoding schemes that result in reduced byte size or include redundancy/error correction and those would certainly be better choices if you don't care about streaming data.

For our case, when we use cobs::encode on the microcontroller, we will _always_ append a marker byte to the end of the buffer as the encoded data will then be passed to a low level driver that actually handles the transmission.

Unless there is a specific use case that benefits from excluding the marker byte, we should append it as that is how the algorithm is normally described and implemented.

inc/cobs.h
oliverlee: EDIT: Actually, we don't know the size of encoded data until it's encoded right? (For input data that's more than 254 bytes)

inc/cobs.h
oliverlee: 👍

tests/cobs_test.cc
oliverlee: It isn't. We just need to remember to change it when the test framework is implemented.

tests/cobs_test.cc
oliverlee: Fair point 👍

inc/cobs.h
mickvangelderen: Like you said, for our case it makes the most sense to simply append the marker byte. Hence the decision.

inc/cobs.h
mickvangelderen: It is, thanks.

inc/cobs.h
mickvangelderen: Nope, we only know the maximum size which is not far off from the actual size.

src/cobs.cc
mickvangelderen: Yes, thanks.

inc/cobs.h
mickvangelderen: An offset of 0 in marker byte 3 means the same as an offset of 1 in marker byte 0: it specifies that 0 data bytes will follow and then a marker byte should be inserted (except if we arrived at the marker byte/end of encoded packet).

`[3, 3, 3]` would be encoded as `[(0), {0}, {0}, {0}, 3]`

oliverlee: After some thought, it seems to me that 0 would denote a offset of 3, as I think the example would be encoded as:
```
[3, 3, 3] -> [1, 1, 1, 1, 3]
```

inc/cobs.h
```diff
@@ -0,0 +1,71 @@
...
+
+3. If the encoding results in { n, ..., x, 0 } where n is an offset and
```
oliverlee: Is n is an offset, where does it point to? This is not clear to me.

src/cobs.cc
```diff
@@ -0,0 +1,124 @@
...
+        debug_assert(dst_copy_idx < dst_len, "buffer overflow");
+        dst[dst_copy_idx++] = 0x00;
+
+        return dst_copy_idx;
+    }
+
+    /*
+    Decodes a consistent overhead byte stuffing encoded buffer.
+    */
+    size_t decode(const uint8_t * const src, size_t src_len, uint8_t * const dst, size_t dst_len) {
```
oliverlee: What about adding a debug assert to check that dst_len is large enough?

src/cobs.cc
```diff
@@ -0,0 +1,124 @@
...
+        debug_assert(encoded_length >= 1, "encoded length must be at least 1");
+        if (encoded_length == 1) return 0;
+        return encoded_length - 2;
+    }
+
+    /*
+    Encodes a byte buffer such that there are no 0 bytes in the output
+    except for the appended 0.
+    */
+    size_t encode(const uint8_t * const src, size_t src_len, uint8_t * const dst, size_t dst_len) {
```
oliverlee: What about adding a debug assert to check that dest_len is large enough?

src/cobs.cc
```diff
@@ -0,0 +1,124 @@
...
+        size_t src_idx = 0;
+        size_t dst_idx = 0;
+        bool write_zero = false;
+
+        debug_assert(src_idx < src_len, "buffer read overflow");
+        uint8_t zero_offset = src[src_idx];
+
+        while (zero_offset != 0x00) {
+            // Compute the index of the next zero.
+            size_t src_zero_idx = src_idx + zero_offset;
```
oliverlee: I think
```cpp
size_t src_zero_idx = src_idx++ + zero_offset;
```
is legible but feel to keep it as is if you don't agree.

oliverlee: This can also be const.

tests/test_cobs.cc
```diff
@@ -0,0 +1,146 @@
...
+Fill a byte buffer with a given value.
+*/
+void fill(uint8_t * const b, size_t l, uint8_t v) {
+    std::memset(b, v, l);
+}
+
+/*
+Check that a byte buffer contains only a given value.
+*/
+void mem_is(uint8_t * const b, size_t l, uint8_t v) {
```
oliverlee: I think length and value would be better argument names.

oliverlee: Renaming this function to `assert_mem_is` would help readability and remind someone else (me) that this function contains an assert.

tests/test_cobs.cc
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+int main() {
+    const uint8_t X = 0xdd; // non-zero payload
+    const uint8_t F = 0xcc; // fill
+
+    // Encode the empty packet.
+    {
+        uint8_t b1[1000]; fill(b1, sizeof(b1), X);
```
oliverlee: encoding/decoding examples as in cobs.h would help here.

tests/test_cobs.cc
```diff
@@ -0,0 +1,146 @@
...
+        uint8_t b2[1000]; fill(b2, sizeof(b2), F);
+        uint8_t b3[1000]; fill(b3, sizeof(b3), F);
+        const size_t b1_len = 0;
+        const size_t b2_len = cobs::encode(b1, b1_len, b2, sizeof(b2));
+        mem_is(&b2[0],   1, 0);
+        mem_is(&b2[1], 999, F);
+        throw_assert(b2_len == 1, "");
+        const size_t b3_len = cobs::decode(b2, b2_len, b3, sizeof(b3));
+        mem_is(&b3[     0],              b1_len, X);
```
oliverlee: This statement doesn't check anything?

oliverlee: Nevermind, I guess I kind of see the point.

tests/test_cobs.cc
```diff
@@ -0,0 +1,146 @@
...
+
+    // Encode the empty packet.
+    {
+        uint8_t b1[1000]; fill(b1, sizeof(b1), X);
+        uint8_t b2[1000]; fill(b2, sizeof(b2), F);
+        uint8_t b3[1000]; fill(b3, sizeof(b3), F);
+        const size_t b1_len = 0;
+        const size_t b2_len = cobs::encode(b1, b1_len, b2, sizeof(b2));
+        mem_is(&b2[0],   1, 0);
+        mem_is(&b2[1], 999, F);
```
oliverlee: seems like ```sizeof(b2) - b2_len``` would be better here in case the buffer size is changed in the future.

tests/test_cobs.cc
```diff
@@ -0,0 +1,146 @@
...
+        mem_is(&b2[255], 745,   F);
+        throw_assert(b2_len == 255, "");
+        const size_t b3_len = cobs::decode(b2, b2_len, b3, sizeof(b3));
+        mem_is(&b3[     0],              b1_len, X);
+        mem_is(&b3[b1_len], sizeof(b3) - b1_len, F);
+        throw_assert(b3_len == b1_len, "");
+    }
+
+    // Encode a packet of 254 non-zero bytes. Result depends on whether or
+    // not the implementation shortens 1|0 to 0.
```
oliverlee: You use the 'shortened' imp now so this should be clearly noted.

oliverlee: You use the 'shortened' imp now so this should be clearly noted.

oliverlee:
A test thing should also be added to travis.

inc/cobs.h
mickvangelderen: That is also a possibility. The table I presented can describe any mapping.

inc/cobs.h
mickvangelderen: The current `decode` implementation decodes both `[ 0 ]` and `[ 1, 0 ]` into `[]`.

src/cobs.cc
mickvangelderen: :+1:

tests/CMakeLists.txt
mickvangelderen: Unlikely.

tests/test_cobs.cc
mickvangelderen: I've changed my mind last night and now think we should use the standard encoding for interoperability and a more elegant implementation of `max_encoded_length` and `max_decoded_length`. I will update the tests afterwards.

mickvangelderen: See previous comment.

oliverlee: 👍

src/cobs.cc
```diff
@@ -2,46 +2,99 @@
...
+    A case where the minimum overhead is achieved is when the decoded
+    byte array consists only of marker bytes. This means that the
+    maximum decoded length can be as big as the encoded length minus
+    the packet overhead, consisting of the first offset and the marker
+    byte. The maximum decoded length is thus given by encoded_length -
+    2.
     */
     size_t max_decoded_length(size_t encoded_length) {
-        debug_assert(encoded_length >= 1, "encoded length must be at least 1");
-        if (encoded_length == 1) return 0;
```
oliverlee: why were these lines removed?

mickvangelderen: I'll push the latest version.

oliverlee:
Why did you change from indexes to pointers? I thought the use of indexes was neat as you didn't need to make a copy of the buffer pointer.

tests/test_cobs.cc
```diff
@@ -1,213 +1,206 @@
...
 */
-void assert_mem_is(uint8_t * buffer, size_t length, uint8_t value) {
-    uint8_t * buffer_end = buffer + length;
-    while (buffer < buffer_end) {
-        throw_assert(*(buffer++) == value, "");
+size_t fill_repetitions(uint8_t * const buffer, std::vector<Rep> repetitions) {
+    size_t offset = 0;
+
+    for (auto const& rep: repetitions) {
+        fill(buffer + offset, rep.count, rep.value);
```
oliverlee: you can resize instead: http://en.cppreference.com/w/cpp/container/vector/resize

mickvangelderen:
@oliverlee change from index to pointers: The compiler will probably keep track of the pointer anyway instead of computing base + offset on every iteation.

tests/test_cobs.cc
oliverlee: Nevermind, I thought you changed buffer to be a vector.

mickvangelderen: Also keep in mind I over allocate on purpose because I want to check if encode/decode writes past the returned length.

inc/cobs.h
```diff
@@ -9,8 +9,29 @@ namespace cobs {
 
     size_t max_decoded_length(size_t encoded_length);
 
-    size_t encode(const uint8_t * const src_start, size_t src_len, uint8_t * const dst_start, size_t dst_len);
+    struct EncodeResult {
+        enum Status {
```
oliverlee: you can change this to enum class to prevent implicit casting to the underlying type
http://en.cppreference.com/w/cpp/language/enum

oliverlee:
Fix commit messages (remove period in commit title, called -> caller).
Change enum to enum class.

mickvangelderen:
Before merging I removed the unused <cmath> header from the first commit and replaced a lingering debug_assert statement with returning a struct in the last commit.

PR #107, opened 2017-02-28
:sparkles: Add editorconfig

mickvangelderen:
The .editorconfig serves as a cross-editor configuration, assuming your editor
has an editorconfig plugin installed and enabled.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+# http://editorconfig.org/
+
+root = true
+
+[{CMakeLists.txt,*.{c,cc,h,hh,py}{,.in}}]
+indent_style = space
+indent_size = 4


.editorconfig
```diff
@@ -0,0 +1,7 @@
+# http://editorconfig.org/
+
+root = true
+
+[*.{cc,h,py}]
```
oliverlee: Should probably add `.c`, `.hh` to that list, along with the optional suffix `.in`
`.in` is used for templates of generated files
and `CMakeLists.txt`

mickvangelderen: :+1:

.editorconfig
```diff
@@ -0,0 +1,7 @@
+# http://editorconfig.org/
+
+root = true
+
+[{CMakeLists.txt,*.{c,cc,h,hh,py}{,.in}}]
```
oliverlee: Does the `{,.in}` mean match nothing OR .in?

mickvangelderen: Jep, its like `(|\.in)` in regex.

mickvangelderen:
@oliverlee Updated the commit message.

PR #112, opened 2017-03-01
Remove SimpleBicycle class

oliverlee:
This class is no longer necessary as equivalent code has been
implemented in BicycleKinematic class of the bicycle submodule.


oliverlee:
mo' code mo' problems

PR #113, opened 2017-03-01
Disable roll angle saturation

oliverlee:
Remove roll angle limit when not using the BicycleKinematic model. Rate
saturation is retained.


oliverlee:
This should have some hardware testing before it gets merged just to make sure the output is sane.

I've left in the rate limits just in case something goes crazy, and also because the state and output normalization functions for the bicycle class only normalize angles and not angular rates.

PR #114, opened 2017-03-01
WIP Fix motor torque calculation

### LINKED ISSUE #109: Missing steer torque term when calculating motor torque
The calculation of motor torque is correct only if the rider steer torque is zero.

oliverlee:
This is a poor way to implement the correct calculation. We need to track across two classes what terms have been calculated and which have not. However, I'm not sure there if there is a better way to implement this. Some other options are to:
~~1. Pass the measured sensor torque to `sim::Bicycle`. This would allow the motor torque calculation to be done entirely in `sim::Bicycle`. However, the steer torque input is calculated in the project file.~~
~~2. Get the steer acceleration from `sim::Bicycle` and do the motor calculation in the project file. This would keep all `physical` terms outside of the `sim::Bicycle` class.~~
3.  Move the handlebar model into `sim::Bicycle` and do 1.

And this certainly needs some hardware tests before merging.

Fixes #109 

- [x] merge #119
- [ ] fix and build and test with hardware

### DIFF
--- a/projects/flimnap/main.cc
+++ b/projects/flimnap/main.cc
@@ -63,6 +63,7 @@ namespace {
                                               sa::RLS_GTS35_ENC_CFG,
                                               MS2ST(1), 3.0f);
     filter::MovingAverage<float, 5> velocity_filter;
+    filter::MovingAverage<float, 5> kistler_filter;
 
     constexpr float fixed_velocity = 5.0f;
 
@@ -264,8 +265,9 @@ int main(void) {
 
         // get sensor measurements
 #if !defined(FLIMNAP_ZERO_INPUT)
-        const float kistler_torque = adc_to_nm(analog.get_adc12(),
-                sa::KISTLER_ADC_ZERO_OFFSET, sa::MAX_KISTLER_TORQUE);
+        const float kistler_torque = kistler_filter.output(
+                adc_to_nm(analog.get_adc12(),
+                    sa::KISTLER_ADC_ZERO_OFFSET, sa::MAX_KISTLER_TORQUE));
 #endif // !defined(FLIMNAP_ZERO_INPUT)
         const float motor_torque = adc_to_nm(analog.get_adc13(),
                 sa::KOLLMORGEN_ADC_ZERO_OFFSET, sa::MAX_KOLLMORGEN_TORQUE);
--- a/projects/flimnap/main.cc
+++ b/projects/flimnap/main.cc
@@ -228,12 +228,6 @@ int main(void) {
     // the Kalman gain matrix.
     bicycle_t bicycle(fixed_velocity, dynamics_period);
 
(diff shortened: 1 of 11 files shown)


oliverlee:
I think option 3 makes the most sense.

projects/flimnap/main.cc
```diff
@@ -275,28 +277,26 @@ int main(void) {
...
+        const float feedback_torque = bicycle.handlebar_feedback_torque();
+        const dacsample_t handlebar_torque_dac = set_handlebar_torque(feedback_torque);
         chTMStopMeasurementX(&computation_time_measurement);
 
         // prepare message for transmission
         msg = msg_init;
         msg.timestamp = starttime;
         message::set_bicycle_input(&msg.input,
-                (model_t::input_t() << roll_torque, steer_torque).finished());
+                (model_t::input_t() << roll_torque, kistler_torque).finished());
```
oliverlee: this is wrong

oliverlee: These values are the input to the Whipple model. As they are calculated in the sim::Bicycle class, we probably need to add a member variable to store the most recent calculation and an accessor.

oliverlee:
I still need to test this on hardware and plot the results.

oliverlee:
Data from two speeds v = 5, v = 7 with commit (Fix sign in motor torque calculation			a7f65a6): [pr114.zip](https://github.com/oliverlee/phobos/files/832009/pr114.zip)

v = 5
screen shot 2017-03-09 at 19 55 21
screen shot 2017-03-09 at 19 55 12

v = 7
screen shot 2017-03-09 at 19 56 50
screen shot 2017-03-09 at 19 57 06

oliverlee:
After adding a 5 element moving average filter:
screen shot 2017-03-09 at 20 18 43
This looks better in that there is no longer a growing oscillation, however, this doesn't stabilize the steering column at all

oliverlee:
Essentially the "scale factor" is a conversion from physical inertia to virtual inertia, assuming it scales linearly. Plots look a lot better though.

screen shot 2017-03-09 at 21 22 47
screen shot 2017-03-09 at 21 23 01

oliverlee:
Steer state plot has poor convergence, likely due to the increase in measurement noise covariance in (WIP Changes to flimnap  … ff87200)

screen shot 2017-03-10 at 10 13 18

oliverlee:
ADC events were not working, resulting in a filter output of zero.

oliverlee:
Okay so this isn't working as it essentially breaks causality.
There are three candidates solutions:
1. Extension of the static handlebar feedback calculation to include roll rate and steer rate terms.
2. Extension of the static handlebar feedback calculation to include roll rate, steer rate, and steer acceleration terms.
3. Configure the motor drive for position control.

oliverlee:
This is invalid.

PR #117, opened 2017-03-02
Change bicycle fixed velocity to 5.0 m/s

oliverlee:
Change the bicycle fixed velocity in Flimnap from 7.0 m/s to 5.0 m/s.
For the Whipple model with benchmark parameters, this is within the
stable range.


oliverlee:
5.0 m/s = 18 kph

5.0 m/s is within the stable speed range for the Whipple model using the benchmark parameters (size of bicycle, mass, etc.). This means that the steer angle (handlebars) will converge to a constant value without any rider input (applying torque to the handlebars).

As a result, things should smash less when testing.

PR #122, opened 2017-03-02
Add observer prime method

### LINKED ISSUE #120: Don't start with a Kalman gain of zero
It takes some time for the state estimate to converge as well as the Kalman gain. We don't want to be sending dumb torque values so use a better initial condition.

oliverlee:
Fixes #120

### DIFF
--- a/projects/flimnap/main.cc
+++ b/projects/flimnap/main.cc
@@ -37,11 +37,11 @@
 #include <array>
 #include <type_traits>
 
-#include "bicycle/kinematic.h" /* simplified bicycle model */
-#include "bicycle/whipple.h" /* whipple bicycle model */
-#include "oracle.h" /* oracle observer */
-#include "kalman.h" /* kalman filter observer */
-#include "haptic.h" /* handlebar feedback */
+#include "bicycle/kinematic.h" // simplified bicycle model
+#include "bicycle/whipple.h" // whipple bicycle model
+#include "oracle.h" // oracle observer
+#include "kalman.h" // kalman filter observer
+#include "haptic.h" // handlebar feedback
 #include "simbicycle.h"
 
 namespace {
@@ -56,7 +56,7 @@ namespace {
 #endif // defined(USE_BICYCLE_KINEMATIC_MODEL)
     using bicycle_t = sim::Bicycle<model_t, observer_t, haptic_t>;
 
-    /* sensors */
+    // sensors
     Analog analog;
     Encoder encoder_steer(sa::RLS_ROLIN_ENC, sa::RLS_ROLIN_ENC_INDEX_CFG);
     EncoderFoaw<float, 32> encoder_rear_wheel(sa::RLS_GTS35_ENC,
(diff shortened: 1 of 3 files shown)


projects/flimnap/main.cc
```diff
@@ -112,22 +112,27 @@ namespace {
     template <typename T>
     struct observer_initializer{
         template <typename S = T>
-        typename std::enable_if<std::is_same<S, observer::Kalman<model_t>>::value, void>::type
-        initialize(S& observer) {
+        typename std::enable_if<std::is_same<typename S::observer_t, observer::Kalman<model_t>>::value, void>::type
+        initialize(S& bicycle) {
+            S::observer_t& observer = bicycle.observer();
```
oliverlee: missing `typename`

projects/flimnap/main.cc
```diff
@@ -112,22 +112,27 @@ namespace {
     template <typename T>
     struct observer_initializer{
         template <typename S = T>
-        typename std::enable_if<std::is_same<S, observer::Kalman<model_t>>::value, void>::type
-        initialize(S& observer) {
+        typename std::enable_if<std::is_same<typename S::observer_t, observer::Kalman<model_t>>::value, void>::type
+        initialize(S& bicycle) {
+            typename S::observer_t& observer = bicycle.observer();
+            observer.set_Q(parameters::defaultvalue::kalman::Q(observer.dt()));
+            /* Reduce steer measurement noise covariance */
```
mickvangelderen: Prefer `//` for comments about blocks of code `/* */` for documenting functions/variable. I suspect the comment also needs to be reflowed/re-wrapped.

projects/flimnap/main.cc
```diff
@@ -112,22 +112,27 @@ namespace {
...
+            observer.set_R(parameters::defaultvalue::kalman::R/1000);
+
+            /* prime the Kalman gain matrix */
```
mickvangelderen: Prefer `//` for comments about blocks of code `/* */` for documenting functions/variable. I suspect the comment also needs to be reflowed/re-wrapped.

projects/flimnap/main.cc
```diff
@@ -112,22 +112,27 @@ namespace {
...
+            bicycle.prime_observer();
+
             /* We start with steer angle equal to the measurement and all other state elements at zero.  */
```
mickvangelderen: Prefer `//` for comments about blocks of code `/* */` for documenting functions/variable. I suspect the comment also needs to be reflowed/re-wrapped.

projects/src/simbicycle.hh
```diff
@@ -161,6 +162,21 @@ void Bicycle<T, U, V>::update_kinematics() {
     m_pose.steer = steer;
 }
 
+template <typename T, typename U, typename V>
```
mickvangelderen: If T, U and V are of a specific .. category, for lack of a better word, then maybe it makes sense to name them less abstractly?

projects/flimnap/main.cc
oliverlee: I've been in consistent with comment type so thanks for bringing this up and I'll make your suggested change.

projects/src/simbicycle.hh
oliverlee: 👍

oliverlee: I would prefer to do this after merging all the pending PRs as there are a number of dependencies.

PR #123, opened 2017-03-02
WIP Increase Kalman error covariance on model change

### LINKED ISSUE #89: Kalman covariances should change when model changes
Our model is parameterized by forward velocity. If the model changes, we expect this change to influence the Kalman gain matrix. I expect the process noise or error covariances should increase at the time the model is changed.

oliverlee:
Fixes #89 

- [ ] merge #122

### DIFF
--- a/projects/flimnap/main.cc
+++ b/projects/flimnap/main.cc
@@ -112,22 +112,27 @@ namespace {
     template <typename T>
     struct observer_initializer{
         template <typename S = T>
-        typename std::enable_if<std::is_same<S, observer::Kalman<model_t>>::value, void>::type
-        initialize(S& observer) {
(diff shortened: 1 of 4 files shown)


projects/flimnap/main.cc
```diff
@@ -291,7 +315,9 @@ int main(void) {
         const float steer_torque = kistler_torque - inertia_torque;
 
         /* simulate bicycle */
-        bicycle.set_v(fixed_velocity);
+        if (bicycle.set_v(fixed_velocity)) {
+            og.model_change_hook();
```
oliverlee: missing arg

oliverlee:
I'm not sure this fix is correct.

oliverlee:
I don't think this is necessary.

PR #129, opened 2017-03-07
Change meaning of COBS result object fields

mickvangelderen:
EncodeResult.written becomes EncodeResult.produced. Its value is only useful
when the status is OK so in other cases it is set to 0.

DecodeResult.read becomes DecodeResult.consumed. Its value is only useful
when the status is OK or UNEXPECTED_ZERO so in other cases it is set to 0. This
makes things slightly easier for code calling `cobs::decode`.

DecodeResult.written becomes DecodeResult.produced. Its value is only useful
when the status is OK so in other cases it is set to 0.

### DIFF
--- a/src/cobs.cc
+++ b/src/cobs.cc
@@ -126,7 +126,7 @@ namespace cobs {
 
         auto bytes_produced = [&dst_copy, &dst_start]() -> size_t {
             // Can cast because dst_copy >= dst_start.
-            return (size_t) (dst_copy - dst_start);
+            return static_cast<size_t> (dst_copy - dst_start);
         };
 
         auto create_ok_status = [&bytes_produced]() -> EncodeResult {
@@ -184,12 +184,12 @@ namespace cobs {
 
         auto bytes_consumed = [&src, &src_start]() -> size_t {
             // Can cast because src >= src_start.
-            return (size_t) (src - src_start);
+            return static_cast<size_t> (src - src_start);
         };
 
         auto bytes_produced = [&dst, &dst_start]() -> size_t {
             // Can cast because dst >= dst_start.
-            return (size_t) (dst - dst_start);
+            return static_cast<size_t> (dst - dst_start);
         };
 
         auto create_ok_status = [&bytes_consumed, &bytes_produced]() -> DecodeResult {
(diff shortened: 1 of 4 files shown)


src/cobs.cc
```diff
@@ -124,21 +124,25 @@ namespace cobs {
         uint8_t * dst_copy = dst_start;
         uint8_t * dst_offset = dst_copy++;
 
-        auto result = [&dst_copy, &dst_start](EncodeResult::Status status) -> EncodeResult {
-            return EncodeResult {
-                status,
+        auto bytes_produced = [&dst_copy, &dst_start]() -> size_t {
+            // Can cast because dst_copy >= dst_start.
+            return (size_t) (dst_copy - dst_start);
```
oliverlee: Please use C++ style casts as they are more explicit with regards to your intention. http://en.cppreference.com/w/cpp/language/static_cast

src/cobs.cc
```diff
@@ -178,27 +182,41 @@ namespace cobs {
         // This variable will always point to the next byte to write.
         uint8_t * dst = dst_start;
 
-        auto result = [&src, &src_start, &dst, &dst_start](DecodeResult::Status status) -> DecodeResult {
-            return DecodeResult {
-                status,
+        auto bytes_consumed = [&src, &src_start]() -> size_t {
+            // Can cast because src >= src_start.
+            return (size_t) (src - src_start);
```
oliverlee: static cast

src/cobs.cc
```diff
@@ -178,27 +182,41 @@ namespace cobs {
...
+        };
+
+        auto bytes_produced = [&dst, &dst_start]() -> size_t {
+            // Can cast because dst >= dst_start.
+            return (size_t) (dst - dst_start);
```
oliverlee: static cast

oliverlee:
Everything looks good except the casts.

src/cobs.cc
```diff
@@ -178,27 +182,41 @@ namespace cobs {
...
-        auto result = [&src, &src_start, &dst, &dst_start](DecodeResult::Status status) -> DecodeResult {
-            return DecodeResult {
-                status,
+        auto bytes_consumed = [&src, &src_start]() -> size_t {
+            // Can cast because src >= src_start.
+            return (size_t) (src - src_start);
+        };
+
+        auto bytes_produced = [&dst, &dst_start]() -> size_t {
+            // Can cast because dst >= dst_start.
```
oliverlee: You can also assert this.

mickvangelderen:
Code is ready, lets fix travis builds first though.

PR #133, opened 2017-03-09
WIP Expand bicycle output state

oliverlee:


oliverlee:
After adding steer rate as a measurement:
screen shot 2017-03-09 at 14 29 49

However this is not the behavior we expect at `v = 5.0 m/s` (stable range):
screen shot 2017-03-09 at 14 29 49

oliverlee:
This is not necessary after implementing admittance control #156

PR #137, opened 2017-03-09
Fix torque adc zero offsets

### LINKED ISSUE #135: Determine offsets for analog input channels
I just realized that we've never verified that what we think is zero is actually zero.

oliverlee:
Add constants defining the ADC value for a corresponding zero torque.
Fix Kistler torque sensor and Kollmorgen actual torque measurements.

Fixes #135


projects/flimnap/main.cc
```diff
@@ -89,6 +89,14 @@ namespace {
         return aout;
     }
 
+    constexpr float convert_adc_to_Nm(adcsample_t value, adcsample_t adc_zero, float magnitude) {
+        // ADC samples are 12 bits
+        // It's not clear when scaling should be applied as data was never saved after the scale
+        // factors were determined.
+        const int16_t shifted_value = static_cast<int16_t>(value) - static_cast<int16_t>(adc_zero);
+        return static_cast<float>(shifted_value)*magnitude/2048;
```
mickvangelderen: Use `2048.0f` in floating point calculations.

projects/flimnap/main.cc
```diff
@@ -89,6 +89,14 @@ namespace {
         return aout;
     }
 
+    constexpr float convert_adc_to_Nm(adcsample_t value, adcsample_t adc_zero, float magnitude) {
```
mickvangelderen: Drop `convert_`, thats clear from the word `to`. I'd consider making it all lower case for consistency. `adc_to_nm` and add a comment about what it does if necessary.

projects/flimnap/main.cc
oliverlee: fixed

projects/flimnap/main.cc
oliverlee: fixed

scripts/plot_sim.py
```diff
@@ -8,8 +8,11 @@
 
 from load_sim import load_messages, get_records_from_messages, get_time_vector
 
+MAX_KISTLER_TORQUE = 50
```
mickvangelderen: Fine for now, creating an issue to deduplicate these values in some way.

mickvangelderen: #139

PR #144, opened 2017-03-09
Add cross editor c++ linting config

mickvangelderen:
.clang_complete is understood by various editors. It provides information that linters and other tools can use.

PR #145, opened 2017-03-09
Enable cobs length computations at compile time

mickvangelderen:
Depends #144


oliverlee:
why is this dependent?

mickvangelderen:
> why is this dependent?

Was building these things one after another and needed them. Will rebase.

PR #150, opened 2017-03-10
Store computed input in sim::Bicycle

oliverlee:
The input vector u is calculated from torque sensor measurements and
inertia properties of the system. This adds a member variable to
sim::Bicycle to store and access the most recently computed input,
allowing code calling the update_dynamics() member function to obtain
the roll torque and steer torque used in model simulation.

### DIFF
--- a/projects/src/simbicycle.hh
+++ b/projects/src/simbicycle.hh
@@ -17,7 +17,8 @@ m_model(v, dt),
 m_observer(m_model),
 m_haptic(m_model, steer_inertia),
 m_state_full(full_state_t::Zero()),
-m_pose(BicyclePoseMessage_init_zero) {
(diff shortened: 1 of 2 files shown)


projects/src/simbicycle.hh
```diff
@@ -112,6 +113,7 @@ void Bicycle<Model, Observer, Haptic>::update_dynamics(real_t roll_torque_input,
         m_observer.set_state(x);
     }
 
+    m_input = input;
```
mickvangelderen: Is there a good reason to create a new `input_t input` and copy it into `m_input` here?

oliverlee: I guess I was worried accessing it from another thread but I'm not doing a good job about it anyways.

PR #155, opened 2017-03-14
Add note about serial device on linux

mickvangelderen:


oliverlee:
I thought this was supposed to be the plugdev group or has that changed for newer versions of ubuntu?

mickvangelderen:
> dialout
> Full and direct access to serial ports. Members of this group can reconfigure the modem, dial anywhere, etc.

from /usr/share/doc/base-passwd/users-and-groups.html on my system.

mickvangelderen:
FYI
> plugdev
> Members of this group can access removable devices in limited ways without explicit configuration in /etc/fstab. This is useful for local users who expect to be able to insert and use CDs, USB drives, and so on.
> 
> Since pmount (the original implementor of group plugdev) always mounts with the nodev and nosuid options and applies other checks, this group is not intended to be root-equivalent in the ways that the ability to mount filesystems might ordinarily allow. Implementors of semantics involving this group should be careful not to allow root-equivalence.

oliverlee:
Cool looks like I've been doing it wrong.

PR #152, opened 2017-03-10
Increase simulation loop rate

oliverlee:
Fixes #147 
- [x] merge #150 
- [x] merge #149 
- [x] test on hardware and confirm no deadline misses: looks good enough to me!

PR #156, opened 2017-03-14
Implement admittance control

### LINKED ISSUE #96: Kalman estimated steer rate does not match numerically calculated steer rate
screen shot 2017-02-24 at 14 33 07

As the encoder is high resolution, a numerically calculated steer rate should be fairly accurate. The above image uses a simple dx/dt calculation. We expect the Kalman filter output should be a closer match. This may be an issue with scaling of the covariance matrices but perhaps modifying the measurement vector of the system from [yaw angle, steer angle] to [yaw angle, steer angle, steer rate] would result in a better estimate in terms of convergence and reduce effort in covariance tuning.

### LINKED ISSUE #104: Scaling difference in state simulation
While oscillatory behavior appears similar, see figures in issue https://github.com/oliverlee/phobos/issues/97, scaling is off by roughly a factor of 4 for rate values

### LINKED ISSUE #109: Missing steer torque term when calculating motor torque

oliverlee:
Fixes #96, fixes #104, fixes #109
- [x] merge #152  
- [x] test on hardware


projects/clustril/main.cc
```diff
@@ -54,11 +56,13 @@ namespace {
...
 
-    dacsample_t set_handlebar_torque(float handlebar_torque) {
-        int32_t saturated_value = (handlebar_torque/sa::MAX_KOLLMORGEN_TORQUE * 2048) + 2048;
-        saturated_value = std::min<int32_t>(std::max<int32_t>(saturated_value, 0), 4096);
-        dacsample_t aout = static_cast<dacsample_t>(saturated_value);
-        dacPutChannelX(sa::KOLLM_DAC, 0, aout);
+    dacsample_t set_handlebar_velocity(float velocity) {
+        // limit velocity to a maximum magnitude of 100 deg/s
+        // input is in units of rad/s
+        const float saturated_velocity = std::min(std::max(velocity, -sa::MAX_KOLLMORGEN_VELOCITY),
```
mickvangelderen: http://en.cppreference.com/w/cpp/algorithm/clamp

projects/clustril/main.cc
```diff
@@ -54,11 +56,13 @@ namespace {
...
+                sa::MAX_KOLLMORGEN_VELOCITY);
+        const dacsample_t aout = saturated_velocity/sa::MAX_KOLLMORGEN_VELOCITY*2048 + 2048;
```
mickvangelderen: What is this magic number 2048?

projects/flimnap/main.cc
```diff
@@ -80,12 +80,13 @@ namespace {
...
-            sa::KOLLMORGEN_DAC_ZERO_OFFSET;
-        saturated_value = std::min<int32_t>(std::max<int32_t>(saturated_value, 0), 4096);
-        dacsample_t aout = static_cast<dacsample_t>(saturated_value);
-        dacPutChannelX(sa::KOLLM_DAC, 0, aout);
+    dacsample_t set_handlebar_velocity(float velocity) {
+        // limit velocity to a maximum magnitude of 100 deg/s
+        // input is in units of rad/s
+        const float saturated_velocity = std::min(std::max(velocity, -sa::MAX_KOLLMORGEN_VELOCITY),
+                sa::MAX_KOLLMORGEN_VELOCITY);
+        const dacsample_t aout = saturated_velocity/sa::MAX_KOLLMORGEN_VELOCITY*2048 + 2048;
```
mickvangelderen: magic number 2048

projects/flimnap/main.cc
```diff
@@ -80,12 +80,13 @@ namespace {
...
-    dacsample_t set_handlebar_torque(float handlebar_torque) {
-        int32_t saturated_value = (handlebar_torque/sa::MAX_KOLLMORGEN_TORQUE * 2048) +
-            sa::KOLLMORGEN_DAC_ZERO_OFFSET;
-        saturated_value = std::min<int32_t>(std::max<int32_t>(saturated_value, 0), 4096);
-        dacsample_t aout = static_cast<dacsample_t>(saturated_value);
-        dacPutChannelX(sa::KOLLM_DAC, 0, aout);
+    dacsample_t set_handlebar_velocity(float velocity) {
+        // limit velocity to a maximum magnitude of 100 deg/s
+        // input is in units of rad/s
+        const float saturated_velocity = std::min(std::max(velocity, -sa::MAX_KOLLMORGEN_VELOCITY),
```
mickvangelderen: http://en.cppreference.com/w/cpp/algorithm/clamp

projects/flimnap/main.cc
```diff
@@ -313,14 +315,14 @@ int main(void) {
...
-        message::set_simulation_actuators(&msg, handlebar_torque_dac);
+        message::set_simulation_actuators(&msg, handlebar_velocity_dac);
         message::set_simulation_sensors(&msg,
                 analog.get_adc12(), analog.get_adc13(),
                 encoder_steer.count(), encoder_rear_wheel.count());
         message::set_simulation_timing(&msg,
                 computation_time_measurement.last, transmission_time_measurement.last);
-        msg.feedback_torque = feedback_torque;
-        msg.has_feedback_torque = true;
+        //msg.feedback_torque = feedback_torque;
```
mickvangelderen: Is there a good reason to leave these comments here?

oliverlee:
This is what it looks like:

screen shot 2017-03-14 at 22 32 26

projects/clustril/main.cc
oliverlee: DAC and ADC are 12-bit. 2048 is half of that.

projects/clustril/main.cc
oliverlee: Cool. Can't use this though. Maybe with the next update of the toolchain.

projects/flimnap/main.cc
oliverlee: no

projects/inc/saconfig.h
```diff
@@ -14,6 +14,9 @@ constexpr const T& clamp(const T& v, const T& lo, const T& hi) {
     return std::min(std::max(v, lo), hi);
 }
 
+constexpr ADC_HALF_RANGE = 4096/2; // ADC is 12-bit
```
mickvangelderen: `(1 << 12)/2` would also have been an option.

oliverlee: i agree

projects/inc/saconfig.h
```diff
@@ -1,10 +1,19 @@
 #pragma once
 #include "hal.h"
 #include "encoder.h"
+#include "debug.h"
+#include <algorithm>
 
 /* sensor and actuator configuration constants */
 namespace sa {
 
+template <typename T>
```
mickvangelderen: Is `saconfig.h` a common/utility file?

oliverlee: The clamp function would probably have a better place elsewhere, but I don't think we have a file for that now unless I forgot about it. saconfig is for sensor and actuator constants.

Issue #130, opened 2017-03-09
Fix COBS implementation


oliverlee:
- [x] rewrite COBS implementation to fix encode/decode of buffers larger than 254 bytes
- [x] replace old (erroneous) COBS encode/decode

mickvangelderen:
Waiting for @oliverlee to review pr

oliverlee:
I can try to look at it on the next few days. No promises since I don't
have a computer.

Issue #77, opened 2017-02-21
Improve plotting after switch to protobuf messages for data transmission


oliverlee:

mickvangelderen:
@oliverlee This seems to have been fixed by you?

oliverlee:
The idea is to have some tool that does a better job of plotting fields and
submessages in a message. The python scripts I have hard-coded the fields I
want to look at.

Ideally this would be written in C++ or something so it loads the messages
faster than python which takes forever.

oliverlee:
Maybe using the hdf5 file format would help as proposed in https://github.com/oliverlee/phobos/issues/186

PR #164, opened 2017-03-16
Replace cobs implementation

mickvangelderen:


projects/flimnap/main.cc
```diff
@@ -66,8 +66,8 @@ namespace {
     constexpr float fixed_velocity = 5.0f;
 
     // transmission
-    std::array<uint8_t, SimulationMessage_size + packet::frame::PACKET_OVERHEAD> encode_buffer;
-    std::array<uint8_t, SimulationMessage_size + packet::frame::PACKET_OVERHEAD> packet_buffer;
+    std::array<uint8_t, SimulationMessage_size> encode_buffer;
```
mickvangelderen: Does this take into account the message size varint?

tools/pbprint.cc
```diff
@@ -3,110 +3,131 @@
...
-#include <asio/serial_port.hpp>
-#include <asio/signal_set.hpp>
 #include <google/protobuf/io/coded_stream.h>
-#include "packet/frame.h"
+#include "cobs.h"
 #include "pose.pb.h"
 #include "simulation.pb.h"
+#include <asio.hpp>
+#include <asio/serial_port.hpp>
+#include <asio/signal_set.hpp>
```
oliverlee: We should come up with a canonincal order for includes at some point.

tools/pbprint.cc
```diff
@@ -116,59 +137,6 @@ namespace {
         }
     }
 
-    void decode_file(const char* filename) {
```
mickvangelderen: Have to update application help or re-add this.

projects/flimnap/main.cc
oliverlee: I don't think it does.

tools/pbprint.cc
oliverlee: 👍

tools/pbprint.cc
```diff
@@ -3,110 +3,131 @@
...
+    void deserialize_packet(const uint8_t * const packet_buffer_start, const size_t packet_buffer_length) {
+        google::protobuf::io::CodedInputStream input(
+            packet_buffer_start,
+            packet_buffer_length
+        );
+
+        uint32_t size;
+        if (!input.ReadVarint32(&size)) {
+            std::cerr << "Unable to decode packet size from serialized data." << std::endl;
+            exit(-1);
```
oliverlee: Please use EXIT_SUCCESS or EXIT_FAILURE unless exit codes are defined somewhere.

tools/pbprint.cc
```diff
@@ -3,110 +3,131 @@
...
+        }
+
+        SimulationMessage msg;
+        if (!msg.ParsePartialFromCodedStream(&input)) {
+            std::cerr << "Unable to parse protobuf message." << std::endl;
+            exit(-1);
+        }
+
+        if (!input.ConsumedEntireMessage()) {
+            std::cerr << "Entire message not consumed. Number of extra bytes: " << input.BytesUntilLimit() << std::endl;
```
oliverlee: You can still print out the decoded message even with this error.

tools/pbprint.cc
```diff
@@ -3,110 +3,131 @@
...
+            // Compute the remaining length.
+            const size_t serial_buffer_len = serial_buffer_write - serial_buffer;
+
+            // Create an output buffer and decode into it.
+            constexpr size_t PACKET_BUFFER_CAPACITY = 2000;
+            uint8_t packet_buffer[PACKET_BUFFER_CAPACITY];
+            cobs::DecodeResult result = cobs::decode(serial_buffer, serial_buffer_len, packet_buffer, PACKET_BUFFER_CAPACITY);
+
+            switch (result.status) {
+            case cobs::DecodeResult::Status::OK: {
```
oliverlee: I prefer indentation on switch/case.

tools/pbprint.cc
```diff
@@ -3,110 +3,131 @@
...
+        uint32_t size;
+        if (!input.ReadVarint32(&size)) {
+            std::cerr << "Unable to decode packet size from serialized data." << std::endl;
+            exit(-1);
+        }
+
+        SimulationMessage msg;
+        if (!msg.ParsePartialFromCodedStream(&input)) {
+            std::cerr << "Unable to parse protobuf message." << std::endl;
+            exit(-1);
```
oliverlee: Please use EXIT_SUCCESS and EXIT_FAILURE.

tools/pbprint.cc
```diff
@@ -3,110 +3,131 @@
...
+        serial_buffer_write += serial_buffer_write_count;
+
+        // Decode as many packets as possible starting from serial_buffer_start.
+        const uint8_t * serial_buffer = serial_buffer_start;
+        while (serial_buffer < serial_buffer_write) {
+            // Compute the remaining length.
+            const size_t serial_buffer_len = serial_buffer_write - serial_buffer;
+
+            // Create an output buffer and decode into it.
+            constexpr size_t PACKET_BUFFER_CAPACITY = 2000;
```
oliverlee: Why not define this near the top of the file where you also defined `SERIAL_BUFFER_CAPACITY`?

.clang_complete
```diff
@@ -1,2 +1,6 @@
+-std=c++1y
 -Iinc
---std=c++14
```
oliverlee: c++14 should be the correct language standard flag.

tools/pbprint.cc
mickvangelderen: I have been using alphabetical order in other projects. Unfortunately sometimes the order is important in c++. We could do something like this when that is the case.

```c
// START ORDERED INCLUDES
#include "c.h"
#include "b.h"
#include "a.h"
// END ORDERED INCLUDES
#include "aa.h"
#include "bb.h"
```

Undid the moved imports.

projects/flimnap/main.cc
mickvangelderen: The varint can be up to 10 bytes according to the [nanopb documentation](https://jpa.kapsi.fi/nanopb/docs/reference.html#pb-encode-varint).

tools/pbprint.cc
mickvangelderen: :+1:

tools/pbprint.cc
mickvangelderen: :+1:

tools/pbprint.cc
mickvangelderen: That is true. However the condition should never evaluate to true so it will only make the code more complex while not adding anything.

tools/pbprint.cc
mickvangelderen: Because it is only used there and I like to keep the definition close to the usage if possible.

tools/pbprint.cc
mickvangelderen: Removed possibility from documenation.

.clang_complete
mickvangelderen: Isn't c++1y the correct language flag?

mickvangelderen:
Fixes #130

.clang_complete
oliverlee: c++1y was used to refer to the c++14 standard before it was completed and implemented in gcc (or whatever compiler). They are functionally equivalent now.

projects/flimnap/main.cc
```diff
@@ -66,10 +66,10 @@ namespace {
     constexpr float fixed_velocity = 5.0f;
 
     // transmission
-    std::array<uint8_t, SimulationMessage_size> encode_buffer;
-    std::array<uint8_t, cobs::max_encoded_length(SimulationMessage_size)> packet_buffer;
+    constexpr std::size_t VARINT_MAX_SIZE = 10;
```
oliverlee: This is defined anywhere by nanopb or by protobuf?

oliverlee:
Go ahead and merge assuming that the varint max size isn't defined by nanopb.

oliverlee:
Also, are you planning on adding an option to decode a file, or do we have to use socat?

projects/flimnap/main.cc
mickvangelderen: Only relevant definition in nanopb is this function: https://github.com/nanopb/nanopb/blob/651bdc45f2180b17c132470ff1a3a515dbffaa78/generator/nanopb_generator.py#L124-L136

Could not find a definition in google's protobuf implementation (https://github.com/google/protobuf/search?l=C%2B%2B&q=varint&type=&utf8=%E2%9C%93), except for this one in a test: https://github.com/google/protobuf/blob/ffa71f80070fde3eb8e6a5779df951c4fe0729e0/conformance/conformance_test.cc

The fact that its defined in this test suggests there is no library definition of the maximum varint byte size.

mickvangelderen:
> Also, are you planning on adding an option to decode a file, or do we have to use socat?

Not in this PR. There is also the open data visualisation issue #77 which might, depending on the chosen solution, render reading serial logs to a text format less important.

Issue #171, opened 2017-04-26
 Add reference to phobos-visualizer in README


oliverlee:
We need to provide a link to the repository with visualizer code, or better yet, offer a prebuilt binary.

PR #180, opened 2017-05-09
Update documentation

### LINKED ISSUE #163: Add tool description to README
There are short descriptions of demos and projects. We should also have short descriptions of tools.

mickvangelderen:
Fixes #163 and #171

### DIFF
--- a/projects/flimnap/README.md
+++ b/projects/flimnap/README.md
@@ -1,0 +1,3 @@
+Runs static simulator code interfacing with Unity
+environment [phobos-visualizer](https://gitlab.com/bikelab/phobos-visualizer).
+Cannot log data via SDIO.
--- a/README.md
+++ b/README.md
@@ -197,6 +197,6 @@ executable is given below:
     +-- projects
         +-- clustril        - Runs development static simulator code. Requires usage of SDIO which prevents usage of serial over USB.
         +-- drunlo          - Prints sensor values as ASCII. Requires usage of serial over USB.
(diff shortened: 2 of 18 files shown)


projects/flimnap/README.md
```diff
@@ -0,0 +1,2 @@
+Runs static simulator code interfacing with Unity environment
+Bikesim(5ade352 or newer). Cannot log data via SDIO.
```
oliverlee: This should point to phobos-visualizer.

oliverlee:
I'm not sure this is the best approach. Before we had a quick summary of the binaries which you could look at the root level README. Now you need to descend down multiple levels to get a brief description.

mickvangelderen:
> I'm not sure this is the best approach. Before we had a quick summary of the binaries which you could look at the root level README. Now you need to descend down multiple levels to get a brief description.

There is a balance I guess. I like localized readmes right now, it encourages exploration of the repository. It also makes it slightly easier to keep documentation up to date because it is closer to the subject. 

The single line descriptions of some of the projects do not do them justice. 

I don't mind taking either approach or a combination.

oliverlee:
I like the idea of having single line descriptions at the root level and
more detailed descriptions in the localized readmes.

On Thu, May 11, 2017 at 9:32 AM, Mick van Gelderen <notifications@github.com

oliverlee:
Opened new issue for improving localized READMEs: https://github.com/oliverlee/phobos/issues/182

Issue #182, opened 2017-05-11
Improve leaf READMEs


oliverlee:
Give better descriptions of demo/project/tool functionality and objectives in the lead README.

This would also be a better place for the matching commit of phobos-visualizer necessary for clustril.

PR #189, opened 2017-05-19
Update readme files

oliverlee:
Fixes https://github.com/oliverlee/phobos/issues/182.

### DIFF
--- a/demos/adc/README.md
+++ b/demos/adc/README.md
@@ -1,1 +1,4 @@
-Reads ADC channels ADC10, ADC11, ADC12.
+This demo reads ADC channels ADC10, ADC11, ADC12. The ADC conversion group is
+[ADC10, ADC11, ADC12] and is read at approximately 1 kHz. The ADC values are
+transmitted over serial in ASCII in units of `adcsample`, a 12-bit unsigned
+value corresponding to a range of 0V to 3.3V.
--- a/demos/bicycle/README.md
+++ b/demos/bicycle/README.md
@@ -1,2 +1,11 @@
-Simulates the Whipple model with Kalman filter and prints state and
-auxiliary state.
+This demo simulates the Whipple bicycle model using the benchmark parameters. A
+Kalman filter is used to compute the state estimate for timing purposes; no
+noise is added to the output of the Whipple model before it is input to the
+Kalman filter. The initial state is set to zero and a periodic pulse (5.0 sec period, 2.0 N-m amplitude, 20% duty cycle) is applied to steer torque.
+
+The simulation runs at roughly 200 Hz. At the end of each simulation loop, the
+following information is transmitted over serial as ASCII:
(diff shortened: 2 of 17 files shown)


demos/adc/README.md
```diff
@@ -1 +1,3 @@
-Reads ADC channels ADC10, ADC11, ADC12.
+This demo reads ADC channels ADC10, ADC11, ADC12. The ADC conversion group is
+[ADC10, ADC11, ADC12] and is read at approximately 1 kHz. The ADC values are
+transmitted over serial in ASCII in units of bits.
```
mickvangelderen: What does "units of bits" mean?

demos/bicycle/README.md
```diff
@@ -1,2 +1,11 @@
...
+Kalman filter is used to compute the state estimate for timing purposes; no
+noise is added to the output of the Whipple model before it is input to the
+Kalman filter. The initial state is set to zero and a periodic pulse (5.0 sec period, 2.0 N-m amplitude, 20% duty cycle) is applied to steer torque.
+
+The simulation runs at roughly 200 Hz. At the end of each simulation loop, the
+following information is transmitted over serial as ASCII:
+ - error between the estimate error for state and auxiliary state in SI units
+ - model state update time in realtime counts
+ - Kalman filter estimate update time in realtime counts
+ - model auxiliary state update in realtime counts
```
mickvangelderen: Perhaps it is more important to describe the goal and the techniques used to achieve it than to give specifics. Unless the documentation is kept up-to-date *very well*, someone who needs the specifics must check the code or put the application to the test anyway.

demos/adc/README.md
oliverlee: The ADC driver we use is 12-bit and converts the voltage range 0V to 3.3V to some digital value between 0 and 4095. ChibiOS calls this a `adcsample` so I'm not sure what the best word to use is.

demos/bicycle/README.md
oliverlee: It was my intention that the local README is kept up to date with the code so you wouldn't need to look at it.

demos/adc/README.md
mickvangelderen: > 12 bit unsigned value with a range of 0V to 3.3V

sounds good.

demos/bicycle/README.md
mickvangelderen: I need a checklist for reviewing then.

oliverlee:
Leaf READMEs should contain:
- [ ] short description of what is done
- [ ] sample rate
- [ ] loop rate (if different than sample rate)
- [ ] transmission rate (if different than sample and loop rate)
- [ ] description of transmitted message
- [ ] list of sensor and drivers that are used.

oliverlee:
@mickvangelderen I think this is done.

PR #191, opened 2017-06-06
WIP Add assistive roll torque to bicycle simulation

### LINKED ISSUE #169: Bicycle simulation stops when velocity is zero
If we are running the simulation and stop pedaling, the velocity of the bicycle goes to zero. This also stops the state of the bicycle from changing as seen here:

```cpp
    if (m_dt == static_cast<real_t>(0)) { // discrete time state does not change
        m_Ad.setIdentity();
        m_Bd.setZero();
    } else {
```
https://github.com/oliverlee/bicycle/blob/0ffe6f2ba4de25b1c01fb78ef64168aaae1deadf/src/bicycle/bicycle.cc#L276-L279

We should allow the simulation to advance in time for zero velocity and determine how to handle special cases (e.g. start of the simulation before the rider starts pedaling).

oliverlee:
Use an LQR controller to generate an assistive virtual roll torque to
stabilize the virtual bicycle at low speed and at large roll angles.
Remove artificial bicycle state limits from sim::Bicycle class.

Fixes #169

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ set(NANOPB_SRC_ROOT_FOLDER ${PROJECT_SOURCE_DIR}/external/nanopb)
 include(GetGitRevisionDescription)
 get_git_head_revision(GIT_REFSPEC GITSHA1)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1z")
 
 ## Define macro for phobos project executable
 # This macro adds common sources for all targets in this project and
@@ -72,6 +72,8 @@ macro(add_phobos_executable target_name)
             if(USE_BICYCLE_EIGEN3_SUBMODULE)
                 set_property(SOURCE ${src} APPEND_STRING PROPERTY COMPILE_FLAGS " -Wno-deprecated")
             endif()
+            # suppress ChibiOS usb serial warning
+            set_property(SOURCE ${src} APPEND_STRING PROPERTY COMPILE_FLAGS " -Wno-maybe-uninitialized")
             math(EXPR source_file_properties_modified
                 "${source_file_properties_modified} + 1")
         endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ macro(add_phobos_executable target_name)
     if(USE_BICYCLE_BOOST_SUBMODULE)
         # suppress boost undef warnings
         set_property(SOURCE ${BICYCLE_SOURCE} APPEND_STRING PROPERTY COMPILE_FLAGS " -Wno-undef")
+        set_property(SOURCE ${ARGN} APPEND_STRING PROPERTY COMPILE_FLAGS " -Wno-misleading-indentation")
     endif()
     if(USE_BICYCLE_EIGEN3_SUBMODULE)
         # Any request from Eigen to allocate memory from heap results in an assertion failure
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,8 +72,6 @@ macro(add_phobos_executable target_name)
             if(USE_BICYCLE_EIGEN3_SUBMODULE)
                 set_property(SOURCE ${src} APPEND_STRING PROPERTY COMPILE_FLAGS " -Wno-deprecated")
             endif()
-            # suppress ChibiOS usb serial warning
-            set_property(SOURCE ${src} APPEND_STRING PROPERTY COMPILE_FLAGS " -Wno-maybe-uninitialized")
             math(EXPR source_file_properties_modified
                 "${source_file_properties_modified} + 1")
         endif()
@@ -91,9 +89,9 @@ macro(add_phobos_executable target_name)
             math(EXPR source_file_properties_modified
                 "${source_file_properties_modified} + 1")
         endif()
-        if(source_file_properties_modified GREATER 2)
-            break()
-        endif()
+        #if(source_file_properties_modified GREATER 2)
+        #    break()
+        #endif()
     endforeach()
 
     if(USE_BICYCLE_BOOST_SUBMODULE)
--- a/projects/flimnap/main.cc
+++ b/projects/flimnap/main.cc
@@ -220,6 +220,8 @@ int main(void) {
              0, 10).finished(),                 // enable steer control
             model_t::state_t::Zero(),           // reference state
             1);                                 // horizon length in iterations
+    // perform value iterations to reach controller steady state
+    controller.perform_value_iteration(2000);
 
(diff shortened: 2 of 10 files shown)


oliverlee:
Bicycle velocity is not getting logged.

oliverlee:
Adding an LQR object increases the binary size so that it exceeds flash size for debug builds.

oliverlee:
We need to add logging of:
1. bicycle velocity
2. LQR gain matrix
3. calculated virtual roll torque (this is already logged though)
to check that if it is working because it does not look like it is.

projects/flimnap/main.cc
```diff
@@ -307,7 +325,15 @@ int main(void) {
 #endif // defined(FLIMNAP_ZERO_INPUT)
 
         // simulate bicycle
-        bicycle.set_v(fixed_velocity);
+        bicycle.set_v(v);
+        // calculate assistive torque
+        const float roll_torque = ((bicycle.v() < assistive_velocity_limit) ||
+                                   (std::abs(model_t::get_state_element(bicycle.observer().state(),
+                                                                        model_t::state_index_t::roll_angle)) >
+                                             assistive_velocity_limit)) ?
```
oliverlee: wrong limit here

oliverlee:
Since the LQR object is too large, we could potentially pre-compute a number of gains and use those.

projects/flimnap/main.cc
```diff
@@ -240,7 +244,22 @@ int main(void) {
...
+             0, 0, 0,   0,   0,                                         // no yaw angle penalty
+             0, 1, 0,   0,   0,                                         // roll angle penalty
+             0, 0, 0,   0,   0,                                         // no steer angle penalty
+             0, 0, 0, 0.1,   0,                                         // small roll rate penalty
+             0, 0, 0,   0, 0.1).finished(),                             // small steer rate penalty
+            (lqr_t::input_cost_t() <<
+             1, 0,                                                      // enable roll torque cost
+             0, 0).finished(),                                          // disable steer control
+            model_t::state_t::Zero(),                                   // reference state
+            static_cast<model::real_t>(MS2ST(10))/CH_CFG_ST_FREQUENCY); // horizon length
```
oliverlee: This conversion is wrong and is zero for the controller doesn't go anything.

oliverlee:
With a single LQR horizon iteration, timing is:
```
timing {
  computation: 137082
  transmission: 82313
}
```
At 168 MHz, corresponding times are:
computation: 0.816 ms
transmission: 0.490 ms

In the current implementation, transmission is blocking so the 1 ms deadline for each simulation iteration is exceeded.

oliverlee:
Current costs result in the following "steady-state":
```
input {
  u: -554.57239
  u: -25.008341
}
pose {
  timestamp: 801797
  x: 0
  y: 0
  yaw: -0.013047734
  roll: 0.4848893
  pitch: 0.31408811
  steer: 0.0018122583
  rear_wheel: 0
}
model {
  v: 0
}
```
simulation is run on a test microcontroller removed from the simulator setup. Encoder signals are zero. Measured torque is large and negative since the sensor is disconnected.

oliverlee:
We can 'prime' the LQR object in the same way with do with the Kalman object, and then directly multiply the state and feedback gain to obtain the roll torque, skipping the LQR update during simulation.

oliverlee:
Steer angle diverges from steer encoder count:
```
timestamp: 5085621
sensors {
  kistler_measured_torque: 2041
  kollmorgen_actual_torque: 2105
  steer_encoder_count: 151881
  rear_wheel_encoder_count: 0
}
actuators {
  kollmorgen_command_velocity: 4096
}
state {
  x: 6.0754795
  x: -3.191052
  x: 4.5113268
  x: -6204.2114
  x: 35776.8
}
input {
  u: -286460.75
  u: 11.434719
}
pose {
  timestamp: 5084986
  x: 0
  y: 0
  yaw: 3.5369272
  roll: -5.3366075
  pitch: -1.570776
  steer: 4.7774968
  rear_wheel: 0
}
model {
  v: 0
}
```

oliverlee:
For current master (https://github.com/oliverlee/phobos/commit/67093df55bd4d345097e98372441efd4ddcec382) has the following elf sizes:
Debug, no LTO
```
oliver@canopus:~/repos/phobos/build$ ~/toolchain/gcc-arm-none-eabi-5_4-2016q3/bin/arm-none-eabi-size projects/flimnap/flimnap_whipple.elf
   text	   data	    bss	    dec	    hex	filename
 925221	   1284	 130288	1056793	 102019	projects/flimnap/flimnap_whipple.elf
```
Release, LTO
```
oliver@canopus:~/repos/phobos/build$ ~/toolchain/gcc-arm-none-eabi-5_4-2016q3/bin/arm-none-eabi-size projects/flimnap/flimnap_whipple.elf
   text	   data	    bss	    dec	    hex	filename
 244148	   1252	 130292	 375692	  5bb8c	projects/flimnap/flimnap_whipple.elf
```

oliverlee:
Building with optimization flag `-Og` instead of `-O0` of commit caee6d4 results in the following binary size:
```
oliver@canopus:~/repos/phobos/build$ ~/toolchain/gcc-arm-none-eabi-5_4-2016q3/bin/arm-none-eabi-size projects/flimnap/flimnap_whipple.elf
   text	   data	    bss	    dec	    hex	filename
 490465	   1284	 130292	 622041	  97dd9	projects/flimnap/flimnap_whipple.elf
```
This appears to be a good tradeoff for now.

oliverlee:
This PR needs to be cleaned up before review and merge.

oliverlee:
Maybe remove LQR controller and just fix roll angle and roll rate to zero below a critical speed.

CMakeLists.txt
```diff
@@ -36,7 +36,9 @@ set(BICYCLE_SOURCE
 set(BICYCLE_INCLUDE_DIR ${BICYCLE_SOURCE_DIR} ${BICYCLE_SOURCE_DIR}/../inc)
 add_definitions("-DBICYCLE_USE_DOUBLE_PRECISION_REAL=false")
 set_property(SOURCE ${BICYCLE_SOURCE} APPEND_STRING PROPERTY COMPILE_FLAGS
-    " -fno-math-errno -fassociative-math -freciprocal-math -fno-signed-zeros -fcx-limited-range")
+    " -fno-math-errno -fassociative-math -freciprocal-math -fcx-limited-range")
+#set_property(SOURCE ${BICYCLE_SOURCE} APPEND_STRING PROPERTY COMPILE_FLAGS
```
mickvangelderen: Perhaps a comment mentioning the flag you took out and why is better than commented duplicate code.

oliverlee: I'm planning on cleaning up this PR and removing most of the commented code. Anything that's commented should have some note next to it.

oliverlee:
This whole PR will probably be rejected in favor of a simpler scheme (zero roll angle and rate) below a critical speed.

oliverlee:
This may work better after merging #220

PR #192, opened 2017-06-07
Fix seriallog gcc build

oliverlee:


tools/seriallog.cc
```diff
@@ -1,4 +1,5 @@
 #include <exception>
+#include <functional>
```
mickvangelderen: What is up with this sudden header include?

oliverlee: It should have always been there but I guess it gets pulled in with clang and gcc versions < 7.

Issue #87, opened 2017-02-22
Kollmorgen torque commands stop working if not receiving USB data


oliverlee:
usbTransmit() blocks (simulation stops running) so sending in a different thread is probably the easiest approach

See here for more information on usbTransmit behavior:
http://www.chibios.com/forum/viewtopic.php?t=2798

oliverlee:
Proposed solution:
* Add semaphore controlling access to a message
* Simulation loop sets message if it can acquire semaphore and then signals or broadcasts an event that the message is ready. If semaphore cannot be acquired, setting of message is skipped.
* Transmission thread waits on event, acquires semaphore, serializes message, releases semaphore, and then transmits buffer.

oliverlee:
Fixed by https://github.com/oliverlee/phobos/commit/68742c142342b10f1c09d8b51fa8b4c91a858409

Issue #181, opened 2017-05-11
Protobuf analysis


mickvangelderen:
## Why are we using protobuf?

We are using protobuf so that we can add fields to messages, even if we already have some existing messages that we still want to be able to use, without having to convert them to the new format.

Protobuf also lets us generate source files for a number of programming languages from a message specification. This means we don't have to port specialized encoding/decoding functions to multiple languages.

Another benefit of protobuf, for which we do not necessarily need protobuf, is that it always encodes numbers as little endian and provides varint encoding.

## What do we want to achieve?

Eventually we want to send messages back and forth between the embedded and desktop application. Different types of messages will be sent at different frequencies. From the perspective of the micro controller we want to report sensor values, simulation state, system configuration (code revision, timer clock frequencies), errors and respond to action or configuration commands. There are probably more uses than the ones I just enumerated.

## Limitations

We do not use dynamic memory allocation on the micro-controller. 

### Message framing

We use Consistent Overhad Byte Stuffing (COBS) to frame messages. COBS takes a specific byte value  (0x00 in our case) and removes it from the message data so it can be used as a marker. 

Skipping over a messages means reading until the next 0x00 byte. This is slower than if we would prepend the length of the message. However when prepending the message length it would become hard to recover reading from the middle of a message. A combination is possible and currently employed.

### Message types

Protobuf does not provide message differentiation. It expects you to know what kind of messages is coming when decoding. This is highlighted in the [documentation](https://developers.google.com/protocol-buffers/docs/techniques#streaming). However

The suggestion they make [here](https://developers.google.com/protocol-buffers/docs/techniques#union) does not seem to apply to protobuf 3. Protobuf encodes the fields of a message by writing the tag of the field (a number) and the value. By creating an enumeration that does exactly that, you are doing things twice unnecessarily.

nanopb has [documentation](https://jpa.kapsi.fi/nanopb/docs/concepts.html#message-framing) on this subject. They suggest specifying a number for each message type which can then be accessed through the generated C header. This message identifying number can be specified as an nanopb option which I assume is not compatible with the C# generator plugin. This means we would have to expose and connect the message type identifiers to message types manually for the C# code and keep it in sync with the nanopb specific protobuf definition options. 

My most recent suggestion was to let protobuf do the work for us. Simply have a "Master" message with all fields optional (proto3 only supports optional fields anyway) defining the different types of messages.

```protobuf
syntax = "proto3";

message Master {
  message Configuration {
    string commit = 1;
[...]

PR #197, opened 2017-06-12
Add serial transmission thread

### LINKED ISSUE #173: Stop adding redundant pose data to messages
Current flimnap code https://github.com/oliverlee/phobos/blob/53addf38980cc3de646b5483e10f8e9a177445a5/projects/flimnap/main.cc
transmits messages with state, sensor, actuator, timing, etc. data in the simulation loop (at 1 kHz). We also add the pose into these messages, which is used for visualization (the others are used for data analysis). However, the pose is only updated in the kinematics loop (at 120 Hz). We are sending redundant the pose more than necessary. This needs to (or should) be addressed.

oliverlee:
This feature branch implements serial USB transmission in a separate thread. Transmission functionality in contained within a class. Other threads are able to queue either SimulationMessage or BicyclePoseMessage objects for asynchronous transmission. The Transmitter class only allows transmission of its own memory managed objects, these can be allocated and freed via class member functions.

Implementation checklist:
- [x] pose thread calculates pose - 67093df
- [x] tx thread running in background - 99d97f5
- [x] post messages from pose thread to tx thread - ad487a2
- [x] post messages from sim thread to tx thread - 1030980
- [x] implement pose mempool - 09dc609
- [x] tx thread tx pose protobuf messages - 50c5170
- [x] implement sim mempool - e8cc032
- [x] tx thread tx pose + sim protobuf messages - e8cc032
- [x] size mem pools - cffe02a
- [x] tx thread class - a3c0c68
- [x] tx thread tx cfg messages at start - a3c0c68
- [x] handle usb disconnect - e9f47d0

Fixes #173 #87


oliverlee:
#181 will need to be implemented in a separate PR.

projects/flimnap/main.cc
```diff
@@ -176,29 +177,40 @@ namespace {
 
     // Transmission thread definitions
     constexpr std::size_t TX_MSG_BUFFER_SIZE = 16;
+    constexpr std::size_t PB_POSE_POOL_SIZE = 4;
     mailbox_t tx_mailbox;
-    std::array<msg_t, TX_MSG_BUFFER_SIZE> tx_msg_buffer;
+    msg_t tx_msg_buffer[TX_MSG_BUFFER_SIZE];
```
mickvangelderen: I'm assuming `msg_t` is pretty much the same type as `void *`?

projects/flimnap/main.cc
```diff
@@ -176,29 +177,40 @@ namespace {
...
 
             int formatted_bytes;
             static char serial_str[128];
             formatted_bytes = chsnprintf(serial_str, sizeof(serial_str)/sizeof(serial_str[0]),
-                    "[tx] current MB used count is %d, last msg: %d\r\n", count, msg);
-            usbTransmit(SDU1.config->usbp, SDU1.config->bulk_in, (const uint8_t*)serial_str, formatted_bytes);
+                    "[tx] current MB used count is %d, last msg: 0x%x\r\n", count, obj);
+
+            // TODO: encode pose to buffer? or should we have a received an encoded pose already?
+            // TODO: How to determine which pool the object belongs to? Compare address of pool?
```
mickvangelderen: What about having multiple mailboxes, just like we have multiple memory pools?

projects/flimnap/main.cc
oliverlee: ChibiOS describes `msg_t` as:
> A message is a variable of type msg_t that is guaranteed to have the same size of and be compatible with (data) pointers (anyway an explicit cast is needed).

and defines it as an integer:
```
oliver@canopus:~/repos/ChibiOS$ rg msg_t | rg typedef | rg ARMCMx | rg GCC | rg "/rt/"
os/rt/ports/ARMCMx/compilers/GCC/chtypes.h:typedef int32_t             msg_t;          /**< Inter-thread message.      */
```

projects/flimnap/main.cc
oliverlee: I thought we talked about that and concluded it wasn't necessary.

projects/flimnap/main.cc
```diff
@@ -176,29 +177,40 @@ namespace {
...
+            if ((reinterpret_cast<BicyclePoseMessage*>(obj) >= pb_pose_buffer) &&
+                (reinterpret_cast<BicyclePoseMessage*>(obj) <= &pb_pose_buffer[PB_POSE_POOL_SIZE]) &&
+                (obj % sizeof(stkalign_t) == 0)) {
```
mickvangelderen: What is the point of this check?

mickvangelderen: I mean, I can see what you're doing but why?

projects/flimnap/main.cc
```diff
@@ -176,29 +177,40 @@ namespace {
...
+                chPoolFree(&pb_pose_pool, reinterpret_cast<void*>(obj));
```
mickvangelderen: doesn't the pool require some sort of locking?

projects/flimnap/main.cc
oliverlee: `obj % sizeof(stkalign_t) == 0)` checks to make sure that the pointer is aligned correctly so the if statement essentially is to determine if `obj` is a pointer to a valid `BicyclePoseMessage`.

I need to distinguish messages coming from the pose thread vs the simulation thread.

projects/flimnap/main.cc
oliverlee: I-class functions (capital I name suffix) can only be called in a locked state. Normal functions can (only) be called outside of a locked state.
```
/**
 * @brief   Releases an object into a memory pool.
 * @pre     The memory pool must be already been initialized.
 * @pre     The freed object must be of the right size for the specified
 *          memory pool.
 * @pre     The object must be properly aligned to contain a pointer to void.
 *
 * @param[in] mp        pointer to a @p memory_pool_t structure
 * @param[in] objp      the pointer to the object to be released
 *
 * @api
 */
void chPoolFree(memory_pool_t *mp, void *objp) {

  chSysLock();
  chPoolFreeI(mp, objp);
  chSysUnlock();
}
```

oliverlee:
commit 50c5170
Output shows pose is not changing so either simulation is not running or we aren't passing the pose objects correctly.
```
timestamp: 0
pose {
  timestamp: 536892712
  x: 0
  y: 0
  yaw: 0
  roll: 0
  pitch: 0
  steer: 0
  rear_wheel: 0
}
timestamp: 9
timestamp: 10
timestamp: 11
timestamp: 12
timestamp: 13
timestamp: 14
timestamp: 15
timestamp: 1
timestamp: 0
pose {
  timestamp: 536892712
  x: 0
  y: 0
  yaw: 0
  roll: 0
  pitch: 0
  steer: 0
  rear_wheel: 0
}
```

oliverlee:
```
timestamp: 0
pose {
  timestamp: 1331110
  x: 30.383261
  y: 25.735729
  yaw: 1.4968215
  roll: 1.7289985
  pitch: 0.20275442
  steer: 0.018067818
  rear_wheel: -147.62869
}
timestamp: 9
timestamp: 10
timestamp: 11
timestamp: 12
timestamp: 13
timestamp: 14
timestamp: 15
timestamp: 1
timestamp: 2
timestamp: 0
pose {
  timestamp: 1331997
  x: 30.386181
  y: 25.775608
  yaw: 1.4982795
  roll: 1.7291274
  pitch: 0.20276827
  steer: 0.018065985
  rear_wheel: -147.76212
}
```

oliverlee:
commit d57c404 results in failed allocation of the simulation memory pool

oliverlee:
```
oliver@canopus:~/repos/phobos/tools/build$ rg pose --after-context 1 test | tail -n30
--
pose {
  timestamp: 14756664
--
pose {
  timestamp: 14757564
--
pose {
  timestamp: 14758457
--
pose {
  timestamp: 14759357
--
pose {
  timestamp: 14760257
--
pose {
  timestamp: 14761155
--
pose {
  timestamp: 14762057
--
pose {
  timestamp: 14762965
--
pose {
  timestamp: 14763864
--
pose {
  timestamp: 14764764
```

```
oliver@canopus:~/repos/phobos/tools/build$ rg "^timestamp: [1-9]" test | tail -n10
timestamp: 14763974
timestamp: 14764074
timestamp: 14764174
timestamp: 14764274
timestamp: 14764374
timestamp: 14764474
timestamp: 14764574
timestamp: 14764674
timestamp: 14764774
timestamp: 14764874
```

oliverlee:
@mickvangelderen This is PR is ready for review. The following will need to be implemented in other PRs:
- transmission of different protobuf message types, with the pre-transmission of a tag
- template variant of this class so we don't need to hardcode message types and pool sizes

projects/inc/transmitter.h
```diff
@@ -0,0 +1,50 @@
...
+#include <array>
+#include "cobs.h"
+#include "ch.h"
+#include "simulation.pb.h"
+
+namespace message {
+class Transmitter {
+    public:
+        Transmitter();
+        void start(tprio_t priority = (NORMALPRIO + 1));
```
mickvangelderen: I think it is better to leave out the default value for the argument. Whoever is using this class should think about the priority.

projects/src/transmitter.cc
```diff
@@ -0,0 +1,157 @@
+#include "transmitter.h"
+#include "hal.h"
+#include "packet/serialize.h"
+#include "usbconfig.h"
+
+// This define can be useful when sizing message mailbox and memory pools
+#define ASSERT_MESSAGE_MEMORY_LIMIT FALSE
```
mickvangelderen: We are already using chDbgAssert which will be stripped on release builds. What is the purpose of this extra layer?

projects/src/transmitter.cc
```diff
@@ -0,0 +1,157 @@
...
+    // Activate the USB driver and then the USB bus pull-up on D+.
+    // Note, a delay is inserted in order to not have to disconnect the cable
+    // after a reset.
+    board_usb_lld_disconnect_bus();   //usbDisconnectBus(serusbcfg.usbp);
+    chThdSleepMilliseconds(1500);
+    usbStart(serusbcfg.usbp, &usbcfg);
+    board_usb_lld_connect_bus();      //usbConnectBus(serusbcfg.usbp);
+
+    // Block USB device until ready
+    while ((SDU1.config->usbp->state != USB_ACTIVE) || (SDU1.state != SDU_READY)) {
```
mickvangelderen: I find that negating the desired state makes the condition easier to read. 
`while (!(SDU1.config->usbp->state == USB_ACTIVE && SDU1.state == SDU_READY)) {`

projects/src/transmitter.cc
```diff
@@ -0,0 +1,157 @@
...
+    if (m_thread != nullptr) {
+        chSysHalt("Synchronization is not implemented and this function is not thread safe");
+    }
+
+    encode_message(reinterpret_cast<msg_t>(msg));
+    transmit_packet();
+    free_simulation_message(msg);
+}
+
+void Transmitter::initialize() {
```
mickvangelderen: If this function is only called from the constructor, just put the code in the constructor.

projects/src/transmitter.cc
```diff
@@ -0,0 +1,157 @@
...
+BicyclePoseMessage* Transmitter::alloc_pose_message() {
+    void* msg = chPoolAlloc(&m_pose_message_pool);
+#if ASSERT_MESSAGE_MEMORY_LIMIT
+    chDbgAssert(msg != nullptr, "Increase transmitter POSE_MESSAGE_POOL_SIZE");
+#endif
+    return static_cast<BicyclePoseMessage*>(msg);
+}
+
+void Transmitter::free_pose_message(BicyclePoseMessage* msg) {
+    chPoolFree(&m_pose_message_pool, reinterpret_cast<void*>(msg));
```
mickvangelderen: `static_cast`

projects/src/transmitter.cc
```diff
@@ -0,0 +1,157 @@
...
+    chPoolLoadArray(&m_pose_message_pool, static_cast<void*>(m_pose_message_buffer), POSE_MESSAGE_POOL_SIZE);
+    chPoolObjectInit(&m_simulation_message_pool, sizeof(SimulationMessage), nullptr);
+    chPoolLoadArray(&m_simulation_message_pool, static_cast<void*>(m_simulation_message_buffer), SIMULATION_MESSAGE_POOL_SIZE);
+}
+
+void Transmitter::encode_message(msg_t msg) {
+    // For now, we send a simulation message but this needs to be fixed later on.
+    // TODO: Presend protobuf tag. See https://github.com/oliverlee/phobos/issues/181#issuecomment-301825244
+    m_bytes_written = 0;
+    if ((msg % sizeof(stkalign_t)) == 0) {
```
mickvangelderen: why do we care?

projects/src/transmitter.cc
```diff
@@ -0,0 +1,157 @@
...
+   chDbgAssert(encode_result.status == cobs::EncodeResult::Status::OK, "Expected encoding to succeed.");
+
+   return encode_result.produced;
+}
+
+void Transmitter::transmitter_thread_function(void* p) {
+    auto self = static_cast<Transmitter*>(p);
+
+    chRegSetThreadName("transmitter");
+    self->initialize();
```
mickvangelderen: Is initialize supposed to be called from both the constructor and here? Am I misreading?

projects/src/transmitter.cc
```diff
@@ -0,0 +1,157 @@
...
+
+    while (!chThdShouldTerminateX()) {
+        msg_t msg = reinterpret_cast<msg_t>(nullptr);
+        chMBFetch(&self->m_message_mailbox, &msg, TIME_INFINITE);
+        self->encode_message(msg);
```
mickvangelderen: `encode_message` and `transmit_message` are called after each other every time. Because they depend on some instance state, I'd consider merging them into 1 function. Even thought they are separate steps conceptually, they are always both performed in the same sequence.

projects/flimnap/main.cc
```diff
@@ -251,32 +196,29 @@ int main(void) {
...
-    message::set_simulation_full_model_observer(&msg, bicycle);
-    size_t bytes_written = write_message_to_encode_buffer(msg);
-
-    // block until ready
-    while ((SDU1.config->usbp->state != USB_ACTIVE) || (SDU1.state != SDU_READY)) {
-        chThdSleepMilliseconds(10);
+    // Initialize USB data transmission
+    message::Transmitter transmitter;
+    {   // Transmit initial message containing gitsha1, model, and observer data.
+        SimulationMessage* msg = transmitter.alloc_simulation_message();
```
oliverlee: This should always succeed as the transmitter thread has not started.

projects/inc/transmitter.h
oliverlee: Okay. As a note, the user will also need to set STM32_USB_OTG_THREAD_PRIO. Maybe it's possible to change with this function call.

projects/src/transmitter.cc
oliverlee: If you're running with debug asserts enabled, you can allow it to drop packets or not drop packets. Release builds should always drop packets.

projects/src/transmitter.cc
oliverlee: Address of memory pool objects must be stack aligned. It the pointer we receive is not stack aligned, it will not point to a valid object in a memory pool

projects/src/transmitter.cc
oliverlee: No forgot to remove here after I changed some stuff.

projects/src/transmitter.cc
oliverlee: I had the same thoughts as well so let's do that.

oliverlee: Actually, there may be a case where we need to transmission fails and we want to try again? Or perhaps we encode as we want to transmit the same packet to two different endpoints? I guess there's some reasoning behind having two separate functions. Unless you feel strongly, I think I'll leave it as is.

oliverlee:
@mickvangelderen I believe all your comments are addressed.

oliverlee:
also need to ensure nothing bad happens if start is called more than once

On Jun 15, 2017 11:59, "Mick van Gelderen" <notifications@github.com> wrote:

*@mickvangelderen* requested changes on this pull request.

I'm impressed with this PR. Worried about Transmitter::initialize.
------------------------------

In projects/flimnap/main.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122148247>:

> @@ -65,19 +64,11 @@ namespace {

     constexpr float fixed_velocity = 5.0f;

-    // transmission
-    constexpr std::size_t VARINT_MAX_SIZE = 10;
-    std::array<uint8_t, SimulationMessage_size + VARINT_MAX_SIZE>
encode_buffer;
-    std::array<uint8_t,
cobs::max_encoded_length(SimulationMessage_size + VARINT_MAX_SIZE)>
packet_buffer;
-    SimulationMessage msg;
-
-    // kinematics loop
-    constexpr systime_t kinematics_loop_time = US2ST(8333); // update
kinematics at 120 Hz
+    // pose calculation loop
+    constexpr systime_t pose_loop_time = US2ST(8333); // update pose at 120 Hz

pose_loop_period
------------------------------

In projects/flimnap/main.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122148437>:

> @@ -65,19 +64,11 @@ namespace {

     constexpr float fixed_velocity = 5.0f;

-    // transmission
-    constexpr std::size_t VARINT_MAX_SIZE = 10;
-    std::array<uint8_t, SimulationMessage_size + VARINT_MAX_SIZE>
encode_buffer;
-    std::array<uint8_t,
cobs::max_encoded_length(SimulationMessage_size + VARINT_MAX_SIZE)>
packet_buffer;
-    SimulationMessage msg;
-
-    // kinematics loop
-    constexpr systime_t kinematics_loop_time = US2ST(8333); // update
kinematics at 120 Hz
+    // pose calculation loop
+    constexpr systime_t pose_loop_time = US2ST(8333); // update pose at 120 Hz

     // dynamics loop
     constexpr systime_t looptime = MS2ST(1); // 1 ms -> 1 kHz

dynamics_loop_period
------------------------------

In projects/flimnap/main.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122150240>:

>          while (true) {
             systime_t starttime = chVTGetSystemTime();
-            update_and_transmit_kinematics(bicycle);
-            systime_t sleeptime = kinematics_loop_time + starttime -
chVTGetSystemTime();
-            if (sleeptime >= kinematics_loop_time) {
-                continue;
+            a->bicycle.update_kinematics();
+            BicyclePoseMessage* msg = a->transmitter.alloc_pose_message();
+            if (msg != nullptr) {
+                *msg = a->bicycle.pose();
+                if (a->transmitter.transmit_async(msg) != MSG_OK) {
+                    a->transmitter.free_pose_message(msg);

Does this mean we discard the pose message when transmission fails? A
comment explaining the rationale behind the action would be appropriate
here.
------------------------------

In projects/flimnap/main.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122152522>:

>          while (true) {
             systime_t starttime = chVTGetSystemTime();
-            update_and_transmit_kinematics(bicycle);
-            systime_t sleeptime = kinematics_loop_time + starttime -
chVTGetSystemTime();
-            if (sleeptime >= kinematics_loop_time) {
-                continue;
+            a->bicycle.update_kinematics();
+            BicyclePoseMessage* msg = a->transmitter.alloc_pose_message();
+            if (msg != nullptr) {
+                *msg = a->bicycle.pose();
+                if (a->transmitter.transmit_async(msg) != MSG_OK) {
+                    a->transmitter.free_pose_message(msg);
+                }
+            }
+
+            systime_t sleeptime = pose_loop_time + starttime -
chVTGetSystemTime();

Lets say the period is 8333 us, and three consecutive loops take 8200 us,
8400 us and 8000 us. The current implementation will shift the transmission
times.

[ 8200 ] sleep 133 [ 8400 ] yield x [ 8000 ] sleep 333 ...

The transmission is shifted by 8400 - 8333 + x = 67 + x

[ 8200 ] sleep 133 [ 8400 ] yield x [ 8000 ] sleep 333 - (67 + x) ...

The latter can be implemented by keeping a loop counter and calculating the
sleep or yield from thread_start_time + counter * period - current_time().
------------------------------

In projects/flimnap/main.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122152576>:

>          while (true) {
             systime_t starttime = chVTGetSystemTime();
-            update_and_transmit_kinematics(bicycle);
-            systime_t sleeptime = kinematics_loop_time + starttime -
chVTGetSystemTime();
-            if (sleeptime >= kinematics_loop_time) {
-                continue;
+            a->bicycle.update_kinematics();
+            BicyclePoseMessage* msg = a->transmitter.alloc_pose_message();
+            if (msg != nullptr) {
+                *msg = a->bicycle.pose();
+                if (a->transmitter.transmit_async(msg) != MSG_OK) {
+                    a->transmitter.free_pose_message(msg);
+                }
+            }
+
+            systime_t sleeptime = pose_loop_time + starttime -
chVTGetSystemTime();

Not sure if worth.
------------------------------

In projects/flimnap/main.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122152820>:

>      chTMObjectInit(&computation_time_measurement);
     chTMObjectInit(&transmission_time_measurement);

-    // Transmit initial message containing gitsha1, model, and observer data.
-    // This initial transmission is blocking.
-    msg = SimulationMessage_init_zero;
-    msg.timestamp = chVTGetSystemTime();
-    message::set_simulation_full_model_observer(&msg, bicycle);
-    size_t bytes_written = write_message_to_encode_buffer(msg);
-
-    // block until ready
-    while ((SDU1.config->usbp->state != USB_ACTIVE) || (SDU1.state !=
SDU_READY)) {
-        chThdSleepMilliseconds(10);
+    // Initialize USB data transmission
+    message::Transmitter transmitter;
+    {   // Transmit initial message containing gitsha1, model, and
observer data.
+        SimulationMessage* msg = transmitter.alloc_simulation_message();

I assume it is assumed the allocation will succeed?
------------------------------

In projects/inc/transmitter.h
<https://github.com/oliverlee/phobos/pull/197#discussion_r122154991>:

> @@ -0,0 +1,50 @@
+#pragma once
+#include <array>
+#include "cobs.h"
+#include "ch.h"
+#include "simulation.pb.h"
+
+namespace message {
+class Transmitter {
+    public:
+        Transmitter();
+        void start(tprio_t priority = (NORMALPRIO + 1));

I think it is better to leave out the default value for the argument.
Whoever is using this class should think about the priority.
------------------------------

In projects/src/transmitter.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122158873>:

> @@ -0,0 +1,157 @@
+#include "transmitter.h"
+#include "hal.h"
+#include "packet/serialize.h"
+#include "usbconfig.h"
+
+// This define can be useful when sizing message mailbox and memory pools
+#define ASSERT_MESSAGE_MEMORY_LIMIT FALSE

We are already using chDbgAssert which will be stripped on release builds.
What is the purpose of this extra layer?
------------------------------

In projects/src/transmitter.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122159593>:

> +    initialize();
+
+    // Initialize a serial-over-USB CDC driver.
+    sduObjectInit(&SDU1);
+    sduStart(&SDU1, &serusbcfg);
+
+    // Activate the USB driver and then the USB bus pull-up on D+.
+    // Note, a delay is inserted in order to not have to disconnect the cable
+    // after a reset.
+    board_usb_lld_disconnect_bus();   //usbDisconnectBus(serusbcfg.usbp);
+    chThdSleepMilliseconds(1500);
+    usbStart(serusbcfg.usbp, &usbcfg);
+    board_usb_lld_connect_bus();      //usbConnectBus(serusbcfg.usbp);
+
+    // Block USB device until ready
+    while ((SDU1.config->usbp->state != USB_ACTIVE) || (SDU1.state !=
SDU_READY)) {

I find that negating the desired state makes the condition easier to read.
while (!(SDU1.config->usbp->state == USB_ACTIVE && SDU1.state ==
SDU_READY)) {
------------------------------

In projects/src/transmitter.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122159934>:

> +    chDbgAssert(status == MSG_OK, "Increase transmitter MAILBOX_SIZE");
+#endif
+    return status;
+}
+
+void Transmitter::transmit(SimulationMessage* msg) {
+    if (m_thread != nullptr) {
+        chSysHalt("Synchronization is not implemented and this
function is not thread safe");
+    }
+
+    encode_message(reinterpret_cast<msg_t>(msg));
+    transmit_packet();
+    free_simulation_message(msg);
+}
+
+void Transmitter::initialize() {

If this function is only called from the constructor, just put the code in
the constructor.
------------------------------

In projects/src/transmitter.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122161984>:

> +void Transmitter::start(tprio_t priority) {
+    m_thread = chThdCreateStatic(m_wa_transmitter_thread,
+            sizeof(m_wa_transmitter_thread), priority,
+            transmitter_thread_function, this);
+}
+
+BicyclePoseMessage* Transmitter::alloc_pose_message() {
+    void* msg = chPoolAlloc(&m_pose_message_pool);
+#if ASSERT_MESSAGE_MEMORY_LIMIT
+    chDbgAssert(msg != nullptr, "Increase transmitter POSE_MESSAGE_POOL_SIZE");
+#endif
+    return static_cast<BicyclePoseMessage*>(msg);
+}
+
+void Transmitter::free_pose_message(BicyclePoseMessage* msg) {
+    chPoolFree(&m_pose_message_pool, reinterpret_cast<void*>(msg));

static_cast
------------------------------

In projects/src/transmitter.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122163663>:

> +    free_simulation_message(msg);
+}
+
+void Transmitter::initialize() {
+    chMBObjectInit(&m_message_mailbox, m_message_mailbox_buffer, MAILBOX_SIZE);
+    chPoolObjectInit(&m_pose_message_pool,
sizeof(BicyclePoseMessage), nullptr);
+    chPoolLoadArray(&m_pose_message_pool,
static_cast<void*>(m_pose_message_buffer), POSE_MESSAGE_POOL_SIZE);
+    chPoolObjectInit(&m_simulation_message_pool,
sizeof(SimulationMessage), nullptr);
+    chPoolLoadArray(&m_simulation_message_pool,
static_cast<void*>(m_simulation_message_buffer),
SIMULATION_MESSAGE_POOL_SIZE);
+}
+
+void Transmitter::encode_message(msg_t msg) {
+    // For now, we send a simulation message but this needs to be
fixed later on.
+    // TODO: Presend protobuf tag. See
https://github.com/oliverlee/phobos/issues/181#issuecomment-301825244
+    m_bytes_written = 0;
+    if ((msg % sizeof(stkalign_t)) == 0) {

why do we care?
------------------------------

In projects/src/transmitter.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122165184>:

> +   const cobs::EncodeResult encode_result = cobs::encode(
+       m_serialize_buffer.data(), serialize_buffer_len,
m_packet_buffer.data(), m_packet_buffer.size());
+
+   // Encoding only fails when the destination buffer is too small, this
+   // should not happen as long as we only encode SimulationMessage objects
+   // because we allocated m_packet_buffer based on its size.
+   chDbgAssert(encode_result.status ==
cobs::EncodeResult::Status::OK, "Expected encoding to succeed.");
+
+   return encode_result.produced;
+}
+
+void Transmitter::transmitter_thread_function(void* p) {
+    auto self = static_cast<Transmitter*>(p);
+
+    chRegSetThreadName("transmitter");
+    self->initialize();

Is initialize supposed to be called from both the constructor and here? Am
I misreading?
------------------------------

In projects/src/transmitter.cc
<https://github.com/oliverlee/phobos/pull/197#discussion_r122165805>:

> +   // because we allocated m_packet_buffer based on its size.
+   chDbgAssert(encode_result.status ==
cobs::EncodeResult::Status::OK, "Expected encoding to succeed.");
+
+   return encode_result.produced;
+}
+
+void Transmitter::transmitter_thread_function(void* p) {
+    auto self = static_cast<Transmitter*>(p);
+
+    chRegSetThreadName("transmitter");
+    self->initialize();
+
+    while (!chThdShouldTerminateX()) {
+        msg_t msg = reinterpret_cast<msg_t>(nullptr);
+        chMBFetch(&self->m_message_mailbox, &msg, TIME_INFINITE);
+        self->encode_message(msg);

encode_message and transmit_message are called after each other every time.
Because they depend on some instance state, I'd consider merging them into
1 function. Even thought they are separate steps conceptually, they are
always both performed in the same sequence.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/oliverlee/phobos/pull/197#pullrequestreview-44235989>,
or mute the thread

projects/flimnap/main.cc
```diff
@@ -65,10 +65,30 @@ namespace {
...
+        systime_t time;
+
+        chSysLock();
+        time = chVTGetSystemTimeX();
+        if (chVTIsTimeWithinX(time, prev, next)) {
+            chThdSleepS(next - time);
+        } else {
+            chSchDoYieldS();
+        }
+        chSysUnlock();
```
mickvangelderen: I assume the `chSysLock()` and `chSysUnlock()` are used so that in between the query of `time` and the sleep, no other work is being done causing a too long sleep? Is the increased accuracy worth the time spent locking and unlocking?

projects/src/transmitter.cc
mickvangelderen: Would it make sense to only check it on debug builds? It should only happen when we make a programming error. The input to the system should never be able to cause such a pointer to appear in the mail box.

projects/src/transmitter.cc
mickvangelderen: If that is the purpose of the flag, I suggest renaming it to something like `OVERWRITE_OLD_PACKETS_TO_PREVENT_OVERFLOW` or something.

mickvangelderen: Maybe not, I don't know. Doesn't matter much.

projects/src/transmitter.cc
mickvangelderen: Perhaps yes. I don't think allowing mistakes now is worth not having to split up a function later.

projects/src/transmitter.cc
oliverlee: Eh, I suck at naming. I doesn't matter too much so I'll merge this PR but feel free to submit a name change PR after.

mickvangelderen: > There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

projects/flimnap/main.cc
oliverlee: The main reason is due to the ChibiOS system state machine: http://chibios.sourceforge.net/html/concepts.html#system_states

Certain functions can only be called in critical zones (S-Locked or I-Locked) those with *S or *I suffix to denote the respective locked state. In order to sleep a thread, the system __always__ needs to be locked. The function variant without the *S suffix `chThdSleep()`  actually locks and unlocks (`os/rt/src/chthreads.c`):
```cpp
void chThdSleep(systime_t time) {

  chSysLock();
  chThdSleepS(time);
  chSysUnlock();
}
```
which then in turn calls (`os/rt/src/chschd.c`):
```cpp
msg_t chSchGoSleepTimeoutS(tstate_t newstate, systime_t time) {

  chDbgCheckClassS();

  if (TIME_INFINITE != time) {
    virtual_timer_t vt;

    chVTDoSetI(&vt, time, wakeup, currp);
    chSchGoSleepS(newstate);
    if (chVTIsArmedI(&vt)) {
      chVTDoResetI(&vt);
    }
  }
  else {
    chSchGoSleepS(newstate);
  }

  return currp->p_u.rdymsg;
}
```
which is arming a virtual timer to wake up the thread after some period. You certainly don't want your thread with a sleep call to get preempted while it's in the process for getting ready for bed. That'd be like someone coming into your room and changing the alarm time, or something.

PR #198, opened 2017-06-15
Disable low speed roll

### LINKED ISSUE #169: Bicycle simulation stops when velocity is zero

oliverlee:
Set roll angle and roll rate to zero below a critical speed of 1 m/s.

Fixes #169

This PR is dependent on the following PRs: #220, #221 
~~This cannot be tested until #208 is fixed.~~

### DIFF
--- a/projects/flimnap/main.cc
+++ b/projects/flimnap/main.cc
@@ -58,12 +58,6 @@ namespace {
                                               MS2ST(1), 3.0f);
     filter::MovingAverage<float, 5> velocity_filter;
 
-    // transmission
-    constexpr std::size_t VARINT_MAX_SIZE = 10;
-    std::array<uint8_t, SimulationMessage_size + VARINT_MAX_SIZE> encode_buffer;
-    std::array<uint8_t, cobs::max_encoded_length(SimulationMessage_size + VARINT_MAX_SIZE)> packet_buffer;
-    SimulationMessage msg;
-
     // pose calculation loop
     constexpr systime_t pose_loop_period = US2ST(8333); // update pose at 120 Hz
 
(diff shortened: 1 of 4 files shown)


oliverlee:
[logs.zip](https://github.com/oliverlee/phobos/files/1091859/logs.zip)

test0.pb.cobs: rider stars at zero velocity
test1.pb.cobs: rider starts at non-zero velocity

oliverlee:
From looking at the test0:
- velocity is not being logged
- handlebar inertia should probably use the true encoder angle and rate instead of steer angle and rate from the bicycle observer
-  roll rate quick grows away from zero at 5 seconds

screen shot 2017-06-21 at 17 31 54

screen shot 2017-06-21 at 17 32 18

oliverlee:
> velocity is not being logged

It is now - ee74639, b60aba6

> handlebar inertia should probably use the true encoder angle and rate instead of steer angle and rate from the bicycle observer

Appears better after fixing the inertia value -  7487f7e


> roll rate quick grows away from zero at 5 seconds

Likely related to incorrect inertia torque calculation

oliverlee:
There is some motor "jerkiness" when we reach the drive speed limit due to saturation:

screen shot 2017-06-22 at 14 24 09
screen shot 2017-06-22 at 14 25 23

oliverlee:
Tested yesterday. Need to decide if this should be merged or https://github.com/oliverlee/phobos/pull/224.

PR #204, opened 2017-06-15
Remove Haptic template argument from sim::Bicycle

### LINKED ISSUE #159: Code cleanup
After changing the motor drive from torque command to velocity command in #156, we need to cleanup unused code.

oliverlee:
Handlebar feedback is independent of the bicycle simulation. The bicycle
simulation class should not compute handlebar feedback torque nor should
it have a handlebar object member variable.

Fixes #159


oliverlee:
Originally, the bicycle simulation class was supposed to take a Haptic object so it would know how to calculate feedback torque.

That calculation has now been removed from the class.

PR #205, opened 2017-06-15
Add simplified haptic calculation

### LINKED ISSUE #161: Implement Arend's approach for handlebar feedback
Detailed more in #114, but repeated here:
1. Extension of the static handlebar feedback calculation to include roll rate and steer rate terms.
2. Extension of the static handlebar feedback calculation to include roll rate, steer rate, and steer acceleration terms.

This has implemented in two parts.
1. Changes to the bicycle model are here: https://github.com/oliverlee/bicycle/commit/6baef62ee9fca43dad8353cdd204a9300fefd88c
2. Changes to the feedback model are here: https://github.com/oliverlee/phobos/pull/205

oliverlee:
implementation checklist:
- [x] simplified model simulation - https://github.com/oliverlee/bicycle/commit/6baef62ee9fca43dad8353cdd204a9300fefd88c
- [x] simplified torque calculation -  ff4cc93
- [x] removal of oracle observer - ~~https://github.com/oliverlee/phobos/pull/213 (not yet merged)~~
- [x] new project flimnap variant
- [x] review after rebasing
- [x] update READMEs

Fixes #161

This PR is dependent on the following: ~~https://github.com/oliverlee/phobos/pull/204~~, ~~https://github.com/oliverlee/phobos/pull/213~~, ~~https://github.com/oliverlee/bicycle/pull/22~~

### DIFF
--- a/projects/flimnap/main.cc
+++ b/projects/flimnap/main.cc
@@ -40,22 +40,28 @@
 
 #if defined(USE_BICYCLE_KINEMATIC_MODEL)
 #include "bicycle/kinematic.h" // simplified bicycle model
-#else // defined(USE_BICYCLE_KINEMATIC_MODEL)
-#include "bicycle/whipple.h" // whipple bicycle model
-#include "kalman.h" // Kalman filter observer
+#elif defined(USE_BICYCLE_AREND_MODEL)
+#include "bicycle/arend.h" // simplified bicycle model
+#else // No Whipple model simplifications
+#include "bicycle/whipple.h"
+#include "kalman.h" // kalman filter observer
 #include "interpolated_lqr.h" // LQR controller
-#endif // defined(USE_BICYCLE_KINEMATIC_MODEL)
+#endif
 
 namespace {
 #if defined(USE_BICYCLE_KINEMATIC_MODEL)
     using model_t = model::BicycleKinematic;
     using observer_t = std::nullptr_t;
-    using haptic_drive_t = haptic::HandlebarStatic;
-#else // defined(USE_BICYCLE_KINEMATIC_MODEL)
+    using haptic_drive_t = haptic::Handlebar0;
+#elif defined(USE_BICYCLE_AREND_MODEL)
+    using model_t = model::BicycleArend;
+    using observer_t = std::nullptr_t;
+    using haptic_drive_t = haptic::Handlebar1;
+#else
     using model_t = model::BicycleWhipple;
     using observer_t = observer::Kalman<model_t>;
     using lqr_t = controller::InterpolatedLqr<model_t>;
-#endif // defined(USE_BICYCLE_KINEMATIC_MODEL)
+#endif
     using bicycle_t = sim::Bicycle<model_t, observer_t>;
 
     // sensors
@@ -213,7 +219,8 @@ int main(void) {
     // the Kalman gain matrix.
     bicycle_t bicycle(0.0, static_cast<model::real_t>(dynamics_loop_period)/CH_CFG_ST_FREQUENCY);
 
-#if defined(USE_BICYCLE_KINEMATIC_MODEL)
+#if defined(USE_BICYCLE_KINEMATIC_MODEL) || defined(USE_BICYCLE_AREND_MODEL)
+    // Initialize handlebar object to calculate motor drive feedback torque
     haptic_drive_t haptic_drive(bicycle.model());
 #else
     // At low speed, we add an assistive roll torque to stabilize the bicycle.
@@ -230,7 +237,7 @@ int main(void) {
 
 #if !defined(FLIMNAP_ZERO_INPUT)
     // Initialize handlebar object to estimate torque due to handlebar inertia.
-    haptic::Handlebar2 handlebar_model(bicycle.model(), sa::UPPER_ASSEMBLY_INERTIA_PHYSICAL);
+    haptic::Handlebar2 handlebar_inertia(bicycle.model(), sa::UPPER_ASSEMBLY_INERTIA_PHYSICAL);
 #endif  // !defined(FLIMNAP_ZERO_INPUT)
 
     observer_initializer<observer_t> oi;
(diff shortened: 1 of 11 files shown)


oliverlee:
Testing of `flimnap-arend` and `flimnap-kinematic` require the motor drive to be reconfigured for torque mode.

oliverlee:
This PR updates the bicycle submodule and, as a result, requires travis cache to be cleared before it can be successfully built.

oliverlee:
This PR needs to be tested on actual hardware and requires a hardware change and a new motor drive configuration.

oliverlee:
This at least runs on the microcontroller without halting.

projects/flimnap/main.cc
```diff
@@ -297,37 +324,58 @@ int main(void) {
...
+
+            bicycle.update_dynamics(u, z);
+#else
 #if !defined(USE_BICYCLE_KINEMATIC_MODEL)
-        if (bicycle.v() < assistance_velocity_limit) {
-            const float interp = bicycle.v() / assistance_velocity_limit;
-            const model_t::input_t u = controller.control_calculate(bicycle.observer().state(), interp);
-            if (assistance_fade_counter < assistance_fade_period) {
-                ++assistance_fade_counter;
+            if (bicycle.v() < assistance_velocity_limit) {
```
mickvangelderen: What about this? I'm not sure `interp` is really worth it. 
```
if (bicycle.v() < assistance_velocity_limit) {
    strength += assistance_strength_increment;
    if (strength > 1.0f) strenght = 1.0f;
}
else {
    strength -= assistance_strength_decrement
    if (strength < 0.0f) strength = 0.0f;
} ;

// apply strength
```

projects/flimnap/main.cc
```diff
@@ -97,11 +116,15 @@ namespace {
...
-        const dacsample_t aout = saturated_velocity/sa::MAX_KOLLMORGEN_VELOCITY*sa::DAC_HALF_RANGE + sa::DAC_HALF_RANGE;
+    dacsample_t set_handlebar_reference(float reference) {
+#if defined(USE_BICYCLE_KINEMATIC_MODEL) || defined(USE_BICYCLE_AREND_MODEL)
+        constexpr float MAX_REF_VALUE = sa::MAX_KOLLMORGEN_TORQUE;
+#else
+        // limit velocity to a maximum magnitude of 100 deg/s, input units of rad/s
+        constexpr float MAX_REF_VALUE = sa::MAX_KOLLMORGEN_VELOCITY;
+#endif
+        const float saturated_reference = util::clamp(reference, -MAX_REF_VALUE, MAX_REF_VALUE);
+        const dacsample_t aout = saturated_reference/MAX_REF_VALUE*sa::DAC_HALF_RANGE + sa::DAC_HALF_RANGE;
```
oliverlee: Yeah. `MAX_REF_VALUE` is expressed in whatever SI unit. `DAC_HALF_RANGE` is in terms of `daccount_t`.

I can replace the `#if` with a check on the observer type.

projects/flimnap/main.cc
oliverlee: We need to interpolate the LQR feedback gain as the feedback matrix is speed dependent.

You can see on line 351 that `interp` is passed to a controller member function.

oliverlee: Also the LQR controller should only be used with the full Whipple model. I need to fix this `#if` clause.

oliverlee: Actually, looks like it's fine. I'll refactor it so it's a bit easier to read though.

oliverlee:
This commit results in the state growing towards infinity:
```
timestamp: 6155886
sensors {
  kistler_measured_torque: 2048
  kollmorgen_actual_torque: 2054
  steer_encoder_count: 8673
  rear_wheel_encoder_count: 0
}
actuators {
  kollmorgen_command_velocity: 4096
}
state {
  x: 0.026852345
  x: 3.8081386e+21
  x: 0
  x: 1.19352e+22
  x: 0
}
input {
  u: 0
  u: 1.8367972e+21
}
model {
  v: 0
}
auxiliary_state {
  x: 0
  x: 0
  x: 0
  x: 0.3141593
}
timing {
  computation: 61759
  transmission: 0
}
```

oliverlee:
This is implemented as described, however the simplified model does not actually work.

projects/flimnap/main.cc
```diff
@@ -301,42 +317,61 @@ int main(void) {
...
+            model_t::set_input_element(u, model_t::input_index_t::roll_torque, roll_torque);
+            model_t::set_input_element(u, model_t::input_index_t::steer_torque, steer_torque);
+
+            model_t::measurement_t z = model_t::measurement_t::Zero();
+            bicycle.model().set_output_element(z, bicycle.model().output_index_t::steer_angle, steer_angle);
+            bicycle.model().set_output_element(z, bicycle.model().output_index_t::steer_rate, steer_rate);
+
+            bicycle.update_dynamics(u, z);
+#else
+            if (bicycle.v() < assistance_velocity_limit) {
```
mickvangelderen: I understand the principle behind `fade` and how it is different from `interp`. I don't understand why both seem to be applied because I am *guessing* that in the control_calculate computation, `K0+ interp*Kd` affects the 'strength' of the assistance, just like `fade` does in `fade * model_t::get_input_element(u,  ...`. 

In the scenario where the rider has been riding above the `assistance_velocity_limit` for some time and suddenly breaks, the `assistance_fade_counter` needs some time to increase meaning no assistance will be applied while the bicycle is extremely unstable. 

What if we just use a function of `bicycle.v() / asistance_velocity_limit` to have a smooth transition instead of the on/off behaviour? That would make the assistance behaviour easier to predict and solve the unstability problem. 

figure_1

oliverlee: Fade is used when assistance is turned on/off so that it is applied/removed gradually. This is to prevent a step command to the motor.

Interp is used to calculate how much control effort should be applied. We calculate the control effort by multiplying a feedback gain matrix  Kwith the state values. The entries of the feedback gain matrix are dependent on velocity and interpolation is used to roughly determine K(v) from K(v=0) and K(v=v_max).

oliverlee:
Log file here: https://gist.github.com/oliverlee/1fd4628f5b1cba9d7bd43aec5611327f
shows roll increasing until exceed 0.1 radians at timestamp 3621813. Initial timestamp is 3406646. 0.1 radians is reached in 2.15 seconds (sysclk @ 100 kHz).

oliverlee:
`$ rg ^state -A6 arend_log | rg -B1 } | rg "x:" > steer_rate_log.txt`: https://gist.github.com/oliverlee/4e29d08e0e708103071ab1bb43a91425
Steer rate filtering could use some improvement.

oliverlee:
This seems to work but it is extremely hard to stabilize the bicycle

PR #213, opened 2017-06-21
Remove Oracle observer

### LINKED ISSUE #231: Reset bicycle simulation full state on reset
Currently, only the observer state is reset on `sim::Bicycle::reset()`. I think we should also reset the full state of the simulator.

oliverlee:
The Oracle observer doesn't do anything so we shouldn't use it.

Fixes #231

### DIFF
--- a/projects/src/simbicycle.hh
+++ b/projects/src/simbicycle.hh
@@ -73,11 +73,11 @@ void Bicycle<Model, Observer>::set_dt(real_t dt)  {
 template <typename Model, typename Observer>
 OBSERVER_FUNCTION(void) Bicycle<Model, Observer>::reset() {
     m_observer.reset();
-    //TODO: reset full state?
+    m_full_state = full_state_t::Zero();
 }
 template <typename Model, typename Observer>
 NULL_OBSERVER_FUNCTION(void) Bicycle<Model, Observer>::reset() {
-    //TODO: reset full state?
+    m_full_state = full_state_t::Zero();
 }
 
 template <typename Model, typename Observer>
@@ -266,7 +266,7 @@ OBSERVER_FUNCTION(typename BICYCLE_TYPE::full_state_t) Bicycle<Model, Observer>:
     m_observer.update_state(m_input, m_measurement);
 
     if (!m_observer.state().allFinite()) {
-        chSysHalt("");
+        chSysHalt("state elements with non finite values");
     }
 
     { // limit allowed bicycle state
@@ -283,6 +283,7 @@ OBSERVER_FUNCTION(typename BICYCLE_TYPE::full_state_t) Bicycle<Model, Observer>:
                 model_t::state_index_t::roll_angle);
         if (std::abs(roll_angle) > constants::pi) {
             // state normalization limits angles to the range [-2*pi, 2*pi]
+            // and here we constrain it further to [-pi, pi]
             roll_angle += std::copysign(constants::two_pi, -1*roll_angle);
         }
 
@@ -299,8 +300,6 @@ OBSERVER_FUNCTION(typename BICYCLE_TYPE::full_state_t) Bicycle<Model, Observer>:
     // convergence of the observer estimate should keep it low.
     // Also, we have no way to correct auxiliary states as there are no sensors
     // to measure them and that's because they are _purely_ virtual.
-    //
-    // TODO: improve this
     return model_t::make_full_state(model_t::get_auxiliary_state_part(state_full),
                                     m_observer.state());
 }
--- a/projects/src/simbicycle.hh
+++ b/projects/src/simbicycle.hh
@@ -284,6 +284,9 @@ OBSERVER_FUNCTION(typename BICYCLE_TYPE::full_state_t) Bicycle<Model, Observer>:
         if (std::abs(roll_angle) > constants::pi) {
             // state normalization limits angles to the range [-2*pi, 2*pi]
             // and here we constrain it further to [-pi, pi]
+            chDbgAssert((roll_angle <= constants::two_pi) &&
+                        (roll_angle >= -constants::two_pi),
+                        "roll angle not model normalized");
             roll_angle += std::copysign(constants::two_pi, -1*roll_angle);
         }
 
--- a/projects/src/simbicycle.hh
+++ b/projects/src/simbicycle.hh
@@ -279,15 +279,15 @@ OBSERVER_FUNCTION(typename BICYCLE_TYPE::full_state_t) Bicycle<Model, Observer>:
             }
         };
 
-        real_t roll_angle = model_t::get_state_element(m_observer.state(),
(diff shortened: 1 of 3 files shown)


projects/src/simbicycle.hh
```diff
@@ -266,4 +256,59 @@ const typename Bicycle<Model, Observer>::full_state_t& Bicycle<Model, Observer>:
...
+    { // limit allowed bicycle state
+        state_t x = m_observer.state();
+
+        auto limit_state_element = [&x](model::Bicycle::state_index_t index, real_t limit) {
+            const real_t value = model::Bicycle::get_state_element(x, index);
+            if (std::abs(value) > limit) {
+                model::Bicycle::set_state_element(x, index, std::copysign(limit, value));
+            }
+        };
+
```
mickvangelderen: For consistency I'd like to see the following implemented for the `roll_angle`. 

```cpp
auto wrap_state_element = [&x](model::bicycle::state_index_t index, real_t min, real_t max) {
    // Assert max > min
    const real_t value = model::Bicycle::get_state_element(x, index);
    const real_t range = max - min;
    while (value >= max) { model::bicycle::set_state_element(x, index, value - range); }
    while (value < min) { model::bicycle::set_state_element(x, index, value + range); }
}
```

If the value can not change more than `range`, or if we don't care that the value is not guaranteed to be in the specified range,  the `while` loops can be replaced by `if` statements. If the value may change many multiples of `range`, a calculation involving division might be better than the loops.

projects/src/simbicycle.hh
```diff
@@ -266,4 +256,59 @@ const typename Bicycle<Model, Observer>::full_state_t& Bicycle<Model, Observer>:
...
+        real_t roll_angle = model_t::get_state_element(m_observer.state(),
+                model_t::state_index_t::roll_angle);
+        if (std::abs(roll_angle) > constants::pi) {
+            // state normalization limits angles to the range [-2*pi, 2*pi]
+            roll_angle += std::copysign(constants::two_pi, -1*roll_angle);
```
mickvangelderen: The roll_angle does not seem to be updated in the state. Is this intentional?

projects/src/simbicycle.hh
```diff
@@ -266,4 +256,59 @@ const typename Bicycle<Model, Observer>::full_state_t& Bicycle<Model, Observer>:
...
+            const real_t value = model::Bicycle::get_state_element(x, index);
+            if (std::abs(value) > limit) {
+                model::Bicycle::set_state_element(x, index, std::copysign(limit, value));
+            }
+        };
+
+        real_t roll_angle = model_t::get_state_element(m_observer.state(),
+                model_t::state_index_t::roll_angle);
+        if (std::abs(roll_angle) > constants::pi) {
+            // state normalization limits angles to the range [-2*pi, 2*pi]
```
mickvangelderen: Comment should read
```
// Limit roll angle to the range (-pi, pi) (exclusive)
```
https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals

projects/src/simbicycle.hh
```diff
@@ -266,4 +256,59 @@ const typename Bicycle<Model, Observer>::full_state_t& Bicycle<Model, Observer>:
...
+
+    // Merge observer and model states
+    //
+    // We simply copy the observer state estimate to the full state vector
+    // this may result in accumulated error in the auxiliary states but
+    // convergence of the observer estimate should keep it low.
+    // Also, we have no way to correct auxiliary states as there are no sensors
+    // to measure them and that's because they are _purely_ virtual.
+    //
+    // TODO: improve this
```
mickvangelderen: Is there an issue for this todo?

projects/src/simbicycle.hh
```diff
@@ -48,8 +71,13 @@ void Bicycle<Model, Observer>::set_dt(real_t dt)  {
 }
 
 template <typename Model, typename Observer>
-void Bicycle<Model, Observer>::reset() {
+OBSERVER_FUNCTION(void) Bicycle<Model, Observer>::reset() {
     m_observer.reset();
+    //TODO: reset full state?
```
oliverlee: https://github.com/oliverlee/phobos/issues/231

projects/src/simbicycle.hh
```diff
@@ -48,8 +71,13 @@ void Bicycle<Model, Observer>::set_dt(real_t dt)  {
...
+}
+template <typename Model, typename Observer>
+NULL_OBSERVER_FUNCTION(void) Bicycle<Model, Observer>::reset() {
+    //TODO: reset full state?
```
oliverlee: #231

projects/src/simbicycle.hh
```diff
@@ -146,35 +133,38 @@ void Bicycle<Model, Observer>::update_kinematics() {
...
-        state_t x = state_t::Zero();
-        model::Bicycle::set_state_element(x, model_t::state_index_t::steer_angle, 0.1f); // in radians
-
-        float t = 0;
-        while (t < observer_prime_period) {
-            x = m_model.update_state(x);
-            m_observer.update_state(input_t::Zero(), m_model.calculate_output(x));
-            t += m_model.dt();
-        }
+OBSERVER_FUNCTION(void) Bicycle<Model, Observer>::prime_observer() {
```
oliverlee: T is used force template argument deduction
https://stackoverflow.com/questions/30953248/why-doesnt-sfinae-enable-if-work-for-member-functions-of-a-class-template
And details of deduction are here:
http://en.cppreference.com/w/cpp/language/template_argument_deduction
although I haven't read through it throughly.

projects/src/simbicycle.hh
oliverlee: The bicycle class normalizes the angle by using std::fmod(x, 2*pi). Normalizing to the range [-pi, pi) would require two extra additions.

oliverlee: So before this line, the angle is in the range [-2*pi, 2*pi]. This line is supposed to change the range to [-pi, pi].

projects/src/simbicycle.hh
oliverlee: It happens below on line 304.

projects/src/simbicycle.hh
oliverlee: https://github.com/oliverlee/phobos/issues/232

projects/src/simbicycle.hh
oliverlee: Debug assert now added. Hopefully, what is being done is clear.

projects/src/simbicycle.hh
oliverlee: No wait, you're right.

projects/src/simbicycle.hh
mickvangelderen: You're right the bounds are exclusive.

PR #220, opened 2017-06-22
Fix handlebar inertia value

oliverlee:
Use the measured physical upper assembly inertia value for calculation
of upper assembly inertia torque.


oliverlee:
This has been tested here: https://github.com/oliverlee/phobos/pull/198#issuecomment-310365293

PR #221, opened 2017-06-22
Log bicycle velocity to simulation messages

oliverlee:


oliverlee:
This has been tested here: #198 (comment)

PR #240, opened 2017-06-28
Add Kollmorgen motor drive torque mode config

oliverlee:
Add Kollmorgen workbench config file for servo motor AKM42G-ACCNDA00 and
servo drive AKD-P00606-NBEI-E000, with autotune generated parameters for
torque mode.

Fixes https://github.com/oliverlee/phobos/issues/237


oliverlee:
The current motor config does not work. Do not merge yet.

PR #256, opened 2017-07-02
Update Kollmorgen velocity mode parameters

### LINKED ISSUE #253: Scale down motor torque for Whipple model
We need to reduce the torque for the demo. It may possible to use a less aggressive controller in the motor drive instead of modifying the velocity reference.

oliverlee:
Increase gain margin to 16 db and phase margin to 55 deg to lower
bandwidth, decrease noise, reduce system stiffness.

Fixes #253.


oliverlee:
Tested this and it seems to work better. You can actually steer the bicycle easily now.
It is more obvious when the model is in the weave and capsize modes based on simulator behavior.

PR #262, opened 2017-07-12
Add project limtoc

oliverlee:
- [x] add torque mode binary
  - compare actual period vs expected period
- [x] remove hardcoded DAC channel
~~- [ ] add velocity mode binary~~
- [x] rebase from #271 
- [x] remove/replace log files after fixing torque calibration


oliverlee:
Running velocity mode binary shows steer rate grow when steer angle is constant and no torque is applied.

projects/limtoc/main.cc
```diff
@@ -0,0 +1,223 @@
...
+    Encoder encoder_steer(sa::RLS_ROLIN_ENC, sa::RLS_ROLIN_ENC_INDEX_CFG);
+
+    dacsample_t set_handlebar_reference(float reference) {
+#if defined(LIMTOC_VELOCITY_MODE)
+        constexpr float MAX_REF_VALUE = sa::MAX_KOLLMORGEN_VELOCITY;
+#else // defined(LIMTOC_VELOCITY_MODE)
+        constexpr float MAX_REF_VALUE = sa::MAX_KOLLMORGEN_TORQUE;
+#endif // defined(LIMTOC_VELOCITY_MODE)
+        const float saturated_reference = util::clamp(reference, -MAX_REF_VALUE, MAX_REF_VALUE);
+        const dacsample_t aout = saturated_reference/MAX_REF_VALUE*sa::DAC_HALF_RANGE + sa::DAC_HALF_RANGE;
```
mickvangelderen: `adc_to_Nm` is also a map:

```cpp
template <typename T>
constexpr T adc_to_Nm(adcsample_t value, adcsample_t adc_zero, T magnitude) {
    // Convert torque from ADC samples to Nm.
    // ADC samples are 12 bits.
    // It's not clear when scaling should be applied as data was never saved after the scale
    // factors were determined.
    return map(static_cast<T>(value), static_cast<T>(adc_zero), magnitude, T {}, static_cast<T>(sa::ADC_HALF_RANGE));
}

P.S. why was the shift computed in `int16_t`?

Comment (oliverlee) [inline projects/limtoc/main.cc]: `constexpr` is a good idea.

`adcsample_t` is a typedef-ed as `uint16_t` but doesn't use the whole range as the ADCs are 12-bit.

File: projects/limtoc/main.cc [hunk 2]
```diff
@@ -0,0 +1,223 @@
...
+        dacPutChannelX(sa::KOLLM_DAC, 0, aout); // TODO: don't hardcode zero but find the DAC channel constant
```
oliverlee: This is fixed in one of the other PRs.

projects/limtoc/main.cc
```diff
@@ -0,0 +1,223 @@
...
+#include "utility.h"
+
+#include "parameters.h"
+
+#include "discrete_linear.h"
+#include "kalman.h"
+
+namespace {
+    constexpr systime_t loop_period = MS2ST(1);
+    constexpr float dt = static_cast<float>(ST2MS(loop_period))/1000.0f; // seconds
```
oliverlee: Can't use `ST2S` because that returns an int which will be zero.

oliverlee: Or 1. Can't remember the exact conversion formula.

mickvangelderen: Ah you're right. Computes in uint32 and ceils to 1.

projects/limtoc/main.cc
```diff
@@ -190,30 +306,61 @@ int main(void) {
...
 
 #if defined(LIMTOC_VELOCITY_MODE)
-        printf("[%u] kistler: %10.5f Nm\tkollmorgen: %10.5f Nm\tsteer: %10.5f rad\trate: %10.5f rad/s\r\n",
-               chSysGetRealtimeCounterX(), kistler_torque, motor_torque, observer.x()[0], observer.x()[1]);
+        printf("%u, %10.5f, %10.5f, %10.5f, %10.5f, %10.5f, %10.5f\r\n",
+               chSysGetRealtimeCounterX(), kistler_torque, motor_torque, feedback_reference,
+               kalman_encoder.x()[0], polyfit_velocity, polyfit_acceleration);
+
+        // FIXME copy values for now and implement a circular buffer later
+        std::memcpy(velocity_samples.data() + 1,
```
mickvangelderen: Undefined behaviour, std::memmove. 
> The memory areas must not overlap. Use memmove(3) if the memory areas do overlap.
https://linux.die.net/man/3/memcpy

oliverlee: Thanks!

oliverlee:
I am going to split this PR into two: one that implements the torque mode binary and a second that implements velocity mode.

oliverlee:
@mickvangelderen All logs are compressed now.

projects/limtoc/main.cc
```diff
@@ -0,0 +1,153 @@
...
+
+    // sensors
+    Analog analog;
+    Encoder encoder_steer(sa::RLS_ROLIN_ENC, sa::RLS_ROLIN_ENC_INDEX_CFG);
+
+    dacsample_t set_handlebar_reference(float reference) {
+        constexpr float MAX_REF_VALUE = sa::MAX_KOLLMORGEN_TORQUE;
+        const float saturated_reference = util::clamp(reference, -MAX_REF_VALUE, MAX_REF_VALUE);
+        const dacsample_t aout = saturated_reference/MAX_REF_VALUE*sa::DAC_HALF_RANGE + sa::DAC_HALF_RANGE;
+        dacPutChannelX(sa::KOLLM_DAC, 0, aout); // TODO: don't hardcode zero but find the DAC channel constant
```
mickvangelderen: Tech debt issue?

oliverlee: This is implemented correctly here: https://github.com/oliverlee/phobos/commit/258cb7027540e55514bcefb9ba71ad5539b6a191

I'll fix it when I rebase off master.

PR #266, opened 2017-07-26
Fix DAC channel constant

oliverlee:


projects/flimnap/main.cc
```diff
@@ -99,12 +99,17 @@ namespace {
...
+        // Calculate channel value based on DAC device params
+        // regshift = 0 -> CH1 -> channel value 0
+        // regshift = 16 -> CH2 -> channel value 1
+        static const dacchannel_t channel = sa::KOLLM_DAC->params->regshift/16;
         constexpr float MAX_REF_VALUE = std::is_same<observer_t, std::nullptr_t>::value ?
             sa::MAX_KOLLMORGEN_TORQUE :
             sa::MAX_KOLLMORGEN_VELOCITY;
+
         const float saturated_reference = util::clamp(reference, -MAX_REF_VALUE, MAX_REF_VALUE);
         const dacsample_t aout = saturated_reference/MAX_REF_VALUE*sa::DAC_HALF_RANGE + sa::DAC_HALF_RANGE;
```
mickvangelderen: utilize numeric_map?

oliverlee: It's better to specify the value that corresponds to zero/midpoint directly instead of the range limits. We need a numeric mapping function that carefully handles the midpoint value.

oliverlee: @mickvangelderen This is getting merged. Feel free to open a PR about using numeric_map.

PR #268, opened 2017-08-09
WIP Add ADC gyro sensor

oliverlee:
- [x] use molex connector for +5V and GND
- [x] verify everything works after finishing wiring


oliverlee:
It may be easiest to determine the gyro sensor range using the plotting scripts in this PR: https://github.com/oliverlee/phobos/pull/262

oliverlee:
Currently the sensor is receiving +5V from a lab power supply as I do not have the proper molex connectors.

Issue #269, opened 2017-08-15
Fix motor drive scaling


oliverlee:
The motor drive scaling constants are incorrect. Using the given values and commanding 1.0 N-m motor torque in software shows the following:

<img width="896" alt="screen shot 2017-08-15 at 15 22 51" src="https://user-images.githubusercontent.com/2965031/29317736-a1f37d50-81cd-11e7-88f4-231d716cc9d6.png">

Roughly measured values show that a microcontroller command of 1.0 Nm gets interpreted in the drive (analog input panel) as 1.548 Arms. When multiplied by the torque constant, 0.742 Nm/Arms, this results in a reference torque  of 1.1486 Nm: an error of 15%.

oliverlee:
Analog input scale should change from 1.5 -> 1.306 Arms/V. Voltage is converted from 0-3V3 to ±12 V.

Analog output scale should change from 1.5 -> 1.453 Arms/V. Voltage is converted from ±10V to 0-3V3.

oliverlee:
The DAC offset value needs to be fixed.

oliverlee:
With DAC offset correction, motor drive scale factors are:
analog in - 1.244 Arms/V
analog out - 1.440 Arms/V

PR #271, opened 2017-08-17
Fix torque calibration

oliverlee:
Fixes https://github.com/oliverlee/phobos/issues/269

Still need to:
 - [x] add torque mode drive configuration file


oliverlee:
@mickvangelderen this is ready for review

PR #283, opened 2017-11-13
Add oversampling to Analog class

oliverlee:


oliverlee:
Oversampling can be achieved using methods introduced by this PR: https://github.com/oliverlee/phobos/tree/update-analog-averaging

PR #311, opened 2018-02-12
Define new protobuf messages for 2-way communication

oliverlee:
Define protobuf messages for transmission and receipt of data. Switch to
proto3. Update nanopb repository for improvements to proto3 generation.


projects/proto/simulation.proto
```diff
@@ -1,149 +1,97 @@
...
-    optional SymmetricOutputMatrixMessage measurement_noise_covariance =   4;
-    optional KalmanGainMatrixMessage kalman_gain =                         5;
+message pbActuator {
+    uint32 kollmorgen_command_torque = 1;
 }
 
 // 2x2 matrix
-message SecondOrderMatrixMessage {
+message pbSecondOrderMatrix {
     repeated float m = 1;
```
mickvangelderen: Can decrease the size of all matrix messages by adding the option ```(nanopb).fixed_length = true``` which drops the repeat count value.

projects/proto/simulation.proto
```diff
@@ -1,149 +1,97 @@
...
-    optional SecondOrderMatrixMessage C1 =  4;
-    optional SecondOrderMatrixMessage K0 =  5;
-    optional SecondOrderMatrixMessage K2 =  6;
-    optional StateMatrixMessage A =         7;
-    optional InputMatrixMessage B =         8;
-    optional OutputMatrixMessage C =        9;
-    optional FeedthroughMatrixMessage D =   10;
+message pbModelStateSpace {
+    float v = 1;
+    pbStateMatrix A = 2;
```
mickvangelderen: I understand A, B, C and D are commonly in control theory but since you found such nice names for these object I figured maybe we could use those. Don't care about this one as much because I think both naming schemes are sensible.

projects/proto/simulation.proto
```diff
@@ -1,149 +1,97 @@
...
-
-message FirmwareVersion {
-    required bytes f = 1;
+    pbTiming timing = 8;
 }
 
 // states are yaw, roll, steer, roll rate, steer rate (5)
-message BicycleStateMessage {
+message pbModelState {
     repeated float x = 1;
```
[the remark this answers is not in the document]
oliverlee: This was the case before and why we have a separate options file.

projects/proto/simulation.proto
oliverlee: Need to move v out.

projects/proto/simulation.proto
oliverlee: Looks like it only works for bytes: https://github.com/nanopb/nanopb/issues/260

CMakeLists.txt
```diff
@@ -59,47 +59,51 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y")
...
+            set(USE_BICYCLE_SOURCE TRUE)
+        endif()
+    endforeach()
+
+    foreach(src in ${ARGN})
+        get_filename_component(src_name ${src} NAME)
+
+        # suppress Boost undef warnings
+        if(USE_BICYCLE_BOOST_SUBMODULE AND
+           src_name MATCHES "(main|receiver|txrx_common).cc")
```
oliverlee: `receiver.cc` and `txrx_common.cc` shouldn't be using anything from Boost. Perhaps `type_traits` is getting pulled in from Boost instead of the standard library?

oliverlee: `inc/utility.h` uses `boost::math` which pulls in `boost::type_traits`.

